PR #92, opened 2014-08-25
Added README.md in tutorials/GitHub

thekevinlemon2:


thekevinlemon2:
Hey Mike, I'm still planning on adding more pictures and editing the project, but I'd just like to get some feedback from you. Do you think the general format and direction of it is okay?

mikeizbicki:
The writeup looks reasonably correct, but I'm sorry to say that I think it largely missed the point of the assignment.  It looks like basically just a rewrite of lab1 that's missing a lot of information and the step-by-step walk through.  _If Sally struggles with git and github, how is reading your writeup going to help her any more than just lab1 did by itself?_

When we talked, I thought you said you were going to extend the `README.md` file for lab1.  Specifically, that you were going to draw pictures of the git repo tree after each of the commands get run; and add pictures from github of where all the buttons are on the website.  That would make it easier for Sally to walk through the git tutorial and visualize what each of the commands does.  But your current writeup doesn't really help Sally at all.

thekevinlemon2:
I've updated the file with the pictures you've asked for. It's not complete yet but is that format/idea okay?

mikeizbicki:
Images like this: https://camo.githubusercontent.com/f7fee7569b50641d987992c7c1783fcf7ebda796/687474703a2f2f692e696d6775722e636f6d2f426876594c4c672e706e673f31 are good.  You need to add them to lab1, though, not to this writeup.  Also, you need to do three more things:

1) you should annotate them with a `*` to indicate your current position.

2) you need to format all the pictures consistently; right now, all the images have the same x and y dimensions; this is causing the size of the boxes and arrows to be different depending on how much stuff is in the picture; you need to format the pictures so that the boxes and arrows are all the same size, and the x and y dimensions of the pictures is what varies

3) the images of github are showing up with a lot of aliasing artifacts (for example, the f in fork is missing a few strokes); you need to figure out how to display the images so they look nicer

PR #103, opened 2014-08-28
Hw5 - Syscalls

DanTheTingler:
Made changes needed for 95 points - thanks Mike!


mikeizbicki:
You still have a number of inline code snippets surrounded by things other than backticks.

katherinegallaher:
Can you give me an example of one? I thought I got everything but I was getting confused at the end..

DanTheTingler:
Do we need to put backticks on the numbers as well?

mikeizbicki:
It's not that you missed backticks, it's that you have backticks _and_ something else.  For example there's a (`|`) in there somewhere, and lots of things that look like "`perror`".

As for numbers, I don't really care either way.

PR #105, opened 2014-08-28
Hw5 - Syscalls  

katherinegallaher:
I tried to take out all the extra quotation marks etc so hopefully i got everything


mikeizbicki:
merged.

PR #109, opened 2014-08-29
Git and GitHub Cheat Sheet

brunnaam:


mikeizbicki:
This is very nice overall.  With a few small changes, however, it will be even better:
- It is incorrect that `git push` pushes everything except the tags.  Do a little more research into what exactly it will push and then update the file.
- There's a number of small English mistakes.  Get someone else from the class to proofread the file and help you fix them.

Right now, your grade is an 80.  If you fix both of the above problems before Tuesday next week I'll give you a 100.

PR #112, opened 2014-08-29
Assignment 5: Video Guide

rgosh001:
GitHub, VIM, and Terminal video guide is now available on youtube.com
All information can be found in our README.md


mikeizbicki:
Comments:
- Right now, everything is a header for some reason?  Make the things that shouldn't be headers plain text.
- The videos and text don't feel very integrated to me.  Figure out a way to integrate them better.  If you can't, then it's probably best to split it back up into two separate projects.
- The organization is really weird.  Right now there are vim topics mixed with bash topics mixed with a bunch of other things.  It might be best to create several different markdown files, each one focusing on a different specialty.  Or maybe not.  But you need to think a lot more carefully about how to arrange things.
- I haven't watched all the videos yet.  In the ones I have, however, they've all had small errors.  For example, `ls` does not _return_ all of the output on the screen.  It _displays_ it.  It _return_s either 0 or 1 depending on whether it fails or succeeds.  I'm not sure how big a deal this is yet.  The errors annoy me, but other students are likely to not even notice.  The only problem with that is then they might accidentally internalize the error.
- As a general tip for the writing section: Personal statements about your own experience are much better than broad, sweeping statements about everyone.  For example, this is good: "I was now at the mercy of moving my cursor one line at a time to go anywhere until I learned more about the many helpful tools Vim already had just for traveling through code." And this is bad: "Some ways of moving through the text are a little more complicated but easy once a user gets used to them. "
- Avoid generic statements like (at the end of the indentation section): "could be combined with many commands and not just indenting."  Actually give some examples or take it out.  For example, immediately after talking about `v`, you talk about `d`.  You mention how `d` can be combined with different commands like `$`.  Well, `d` can be combined with visual mode too!  Try to tie the sections together with each other a bit more like this.
- In general, there are a lot of extra sentences that can be taken out of the written stuff without losing any information.
- There are a number of commands in the text that need to be surrounded by backticks.
- The whole site feels like a giant infomercial or spam site.  Always taking about the "secrets of vim" and things like that.  I think you're starting to take it a bit too far.
- I don't like that all the videos use the same image.  Did you get my email about how we would put it on my own website to embed the youtube videos?  We'd also be able to make it default to HD that way.  If you don't want to do this, that's fine, but the images should probably be something appropriate to each topic (or just links).

This project has the potential to be an amazing resource for future students, but right now it is not.  In the current state, I would give everyone 60s.  If you make the changes above, and do a _really_ nice job, there's still potential for a lot of extra credit.  The absolute final due date for this is Tuesday, but please try to get something finished before Saturday sometime so I can give you feedback on it.

If you have any questions about this feedback or anything else, let me know.

rgosh001:
Understood. Thank you Mike!

PR #114, opened 2014-08-29
lwang031  

lwang031:
assn 5 commit "lwang subject"


mikeizbicki:
I just want to say again that I'm very impressed.  Some of these topics are pretty advanced!  Other comments:
- rename your directory so that it doesn't include your name, and is only a title of your project
- it looks like you put your code through a word processor or something.  A lot of the symbols have gotten mangled.  For example: `mv -i “$1″ “$newname”` .  This is not a valid bash command because many of the quotes are weird.  Make sure you can just copy paste anything in your file into bash and have it work.
- there are a number of errors with missing/too many backticks within your paragraphs (see the paragraph just below the code snippet I mentioned above)
- Efficiency has a very specific meaning in computer science.  It means the program uses less resources.  But in your examples of process substitution, you haven't demonstrated that the program uses less resources; you've demonstrated that it uses fewer lines of code.
- your example version control system should be git, not svn :)
- You use `find` in a lot of your example code, but you never explain what it does anywhere.  This is a super useful program, so you should explain it.  Maybe make it the first example.
- I like the pictures throughout.  It gives it a nice informal style.  But the one of Frink has the 9gag logo on it.  Find a picture without someone else's logo.
- In the section on math, you should mention that the `((   ))` syntax is actually just syntactic sugar over the `eval` command.
- Any commands that are on their own line should have that `$` in front of them to indicate that they are being typed directly into the bash shell, and are not part of a file.  Also, you should display the actual output of at least some of the examples.  For example, instead of using comments in the math section, you could do something like:

```
$ echo $((3*37+12))
123

$ echo $((2**16-1)) 
65535

$ echo $((103/10)) 
10

$ echo $((103%10))
3
```
- There are a number of small English mistakes throughout the writeup.  Also, there's a few spots where your explanation is a probably a little too vague.  Get someone else in class to review your writeup and fix these mistakes.

As it stands, your project is a 70.  Fix the things above, and I'll give you a 95.  Read on for some extra credit opportunities.

One of the things I really like about this writeup is that it goes into bash commands that even many experienced programmers won't be familiar with.  I actually learned a couple things reading it!  (I didn't know about `ctrl-u` and `ctrl-y`.)  If we can tune things up a little bit more, I think a lot of people on the internet will enjoy reading this (like maybe thousands).  You would need to make three changes.  First, the explanation about special characters should be an explanation about what the different quotation marks mean in bash; you're already doing this a little, but change the emphasis and mention that backticks execute a program.  For example, `test=`echo foo`` runs the command `echo foo` and sticks stdout into the variable `test`.  Second, when you talk about `ctrl-u`, also mention `ctrl-r`.  Third, the section on math should go into a bit more detail about how to use the `bc` command for floating point.  If you do a _really_ good job on all of that, I could see this project being worth 150 points.  But again, it would have to be really good.

mikeizbicki:
Also, deadline would be next Tuesday, although I might be able to extend it a bit if you want to do all that extra credit.

lwang031:
"all" ?

mikeizbicki:
yep

lwang031:
Dear Mike:

i would like to make an appointment with you for homework regrade and
reading a book for extra credit on Tuesday.

sincerely,

Lin Wang

mikeizbicki:
Does noon work?

lwang031:
yes

PR #118, opened 2014-08-30
second round edit, asking for feedback prior to Tuesday's deadline

rgosh001:
Good Afternoon Mike,
Kyler and I have done a lot of new work to re-vamp the page as your told us to before. I also photoshopped new imaged as our video's thumbnail but our deletion video still needs more time to updates it's image, but the other videos are already updated as of this morning from youtube, If you could give us feedback on the README.md file again, we'd greatly appreciate it. What we did this time is to match the style that Sean has written in, with out text. This way students can also read the description below the video as we'v given them a small taste of what each topic talks about and the commands needed to know. We also redid out 4 videos to shorten but get straight to the point with our viewers. Thank you again Mike. We greatly appreciate your time and feedback.


mikeizbicki:
This is shaping up to be an excellent project.  I can tell you guys are putting in hard work, and it's definitely showing in the final product.

Comments:
- I don't know what you mean by "tags are much safer than branches", and I can't think of any meaning for this that would be correct.
- delete this sentence: "I recently learned what a README.md file was and boy-oh-boy, how do I even make one of these"
- You need to standardize how you refer to vim.  Personally I always write it lowercase, and reword my sentences so that it never comes to be the first word in a sentence so I don't need to capitalize it.  It should only be `vim` if you are specifically talking about running the command.  The same goes for git.
- Some sentences are written in the third person, and some in the first person.  Try to convert everything so that the focus is on how _I_ struggled learning vim, for example, and not how _they_ struggle learning vim.  It's possible to write well using the 3rd person, but it's much harder, and all your 3rd person sentences are boring and awkward.
- Words like "utilize" make you sound pretentious.  You "use" vim, or you "edit in" vim.  Code monkeys utilize things.
- A lot of videos have a weird `_` character to the right of them.
- Replacing is in terminal section, but it should be in the vim section.  Also, there's a lot of commands in there that need backticks.
- Put `perror` in a section all on its own at the end.  It doesn't really connect to the other things you have, but it's still really important.  I'll probably end up moving it from your tutorial file to one of the labs/homeworks after you're all done.
- `ssh`ing into the well server should still be under using the terminal.  When you're done, I'll contact the systems group here to see if they'll put your instructions on to the main cs webpage.  I think it'll be a huge help for beginning cs students.
- I don't know what "Spectacle App" means.  This and the `.vimrc` should be moved under the vim section  Most of the stuff in your conclusion section should probably be moved to the vim section somewhere as well.

rgosh001:
Thank you very much for reviewing this for us again Mike. We'll make more changes and get back to you either late sunday or Monday. I hope you have a good weekend and we look forward to presenting our final version Tuesday.

PR #120, opened 2014-09-01
Third Revision, Feed Back Requested

rgosh001:
Good Afternoon Mike,
We've went through your feedback and made all the changes. I noticed that I uploaded the wrong picture for the deletion video so I changed it but it takes unto 24hrs for youtube to process the change, but other than that, It looks and feels a lot better. We tried to find all the small errors in the readme file and fix them but if you could revise it again and give us more feedback, we'd greatly appreciate and look forward to submitting the final version.

-Rashid & Kyler


mikeizbicki:
- Don't have the title be `README.md`.  Make it something like "The `README` file".  Also, you missed backticks on the first readme in the paragraph.
- Lots of paragraphs have messed up formatting
- The title for vim and vimrc is messed up.  Also, you're inconsistent in your use of `vim` vs. vim.  My recommendation is to just say vim all the time, and only use `vim` when you're explicitly talking about the command to type in.  I don't care what you do though as long as you're consistent.
- `cd` and `mv` commands section has random commands at the beginning
- The intro could still be tightened up a bit.  It's fine to link to the playlist, but don't link to the writeup without the videos.
- The overall structure is better.

kryne001:
I put the structure of the commands for cd and mv at the beginning, for if people just need to find the structure instead of reading the whole paragraph. Should we take that out?

PR #122, opened 2014-09-02
Git and GitHub

brunnaam:


mikeizbicki:
You didn't update the information on `git push`, and there's still a handful of English mistakes.  So I'm giving you an 85 on the assignment, which gives you a B overall in the course.

PR #125, opened 2014-09-02
Fourth Revision, Last Minute Questions

rgosh001:
Good Morning Mike,
Kyler and I just did a lot of reformatting but have a little concern about our intro. We did a little editing to the intro but don't know how to proceed to improving it. We would love for some recommendations on some last minute things to add to the intro if there are some things that should be in there.
Other than that, we decided to keep vim in back-ticks and made sure all of them are consistent.
We also got rid of the link for the write up as it only consisted of Sean's writing because it didn't make sense to us to have a link to the write up if it didn't have Kyler and I's writeup. What I could do is make a separate text file with only the text in there that way it has all of our writeup's in there. Let me know if that's a good idea.
Lastly, we also fixed the formatting with the paragraphs. We didn't realize that a lot of our lines had two extra spaces at the end, which made the next line separate from the paragraph.
Hope you enjoy our project as much as Kyler and I did making it, we will make the next few changes once you reply and we will see you at 12:30pm today.

Thank you Mike!
-Kyler & Rashid


mikeizbicki:
Please read all the comments before you start on the new videos; many of the comments apply to videos other than the video it's listed under.  

One big picture comment is to make all the videos have the same format:  One terminal on the left, one safari window or another terminal on the right, and the keyboard in the bottom right corner.  Also, please figure out how to zoom in on the terminal (`ctrl` `+` will probably do it, if not google).  You should have NO UNUSED WHITESPACE ANYWHERE.  Fill it with something useful for the viewer.

### video comments

getting started:
- movie glitches when you create an account
- you use `pwd` to "see where you're at" not `ls`
- you use the phrase "here" a lot.  Better would be to talk about the button "in the top right corner" than the button "here"

creating a file push/pull
- remove "and so on and so on"
- you did not clone a directory test from your repository last time
- you do NOT run `git init` after you  run `git clone`; you only run one OR the other, never both
- "your repository can go between github and your local machine" doesn't make sense
- don't say "back over here" say "back to the terminal"
- you should explicitly remind people that you're typing `:wq` to save and quit; also say that you're pressing `i` to enter insert mode and `esc` to leave
- your git is not properly configured in all the videos and displays many warnings about how to configure it.  Fix this or no one will take you seriously!!!
- show safari before and after the git push
- you got the "junk file" because you used the `git add -A` command; the correct way to delete a file from a repo is `git rm` not `rm`; you should lookup the `.gitignore` file; github will create one of these for you automatically if you ask it; definitely do NOT talk about `-A` unless you also talk about `.gitignore`
- this seems like a good spot to talk about `git status`
- demonstrate what happens when you run `git commit` without the `-m` flag?  only if you want

branches
- don't say "there's nothing else _there_" say explicitly where
- you "don't add files to your github account," you "add them to your repo"
- the first time you push a branch, you must explicitly say which branch; after that, it will automatically be included in the list of branches to push
- try to be concrete about which specific branches you're talking about; for example, don't say "other branch", say "master branch" or "[branchname] branch"
- you should demonstrate a merge conflict and how to resolve the conflict; in a merge conflict, you WILL have to run the `git add` and `git commit` commands again

tags: 
- I don't understand how tags are more complicated than branches?
- you're not "checking out your second _one_" you're checking out your "second _branch_"; always be specific!
- I like the checkpoint in a game analogy; keep that!
- never put dates in a commit message!  git tracks that for you automatically, if you do it yourself you can only mess things up
- I like using the `--tags` parameter
- don't say "that's it" then have more stuff to talk about!
- You delete tags the same way you delete branches; maybe that's worth mentioning?
- the stuff you're saying about spaces in the command line is just not true, and I don't know where it's coming from
- I recommend reading up on what the `:` means exactly and trying to give a more detailed explanation.  See: http://stackoverflow.com/questions/7303687/why-git-use-the-colon-branch-to-delete-remote-branch and the links inside  (if you don't fully understand, however, then there's no need to put it in the video)

forking
- you don't "fork into a repository" and you don't "do pull requests into a forked repository"; you "issue pull requests from the fork to the original repo"
- you should give a 15 sec motivating example for forking about open source software
- the example repo should be your rshell (something you fully own) and not something you forked from somewhere else; make your username something very different from your `kryne001` so that there's less chance of confusion; also, make a meaningful change, for example, have a typo in your rshell readme and show how to fix the typo and explain that this is how open source software gets better; this will let you actually accept the pull request; commit message should be something simple like "fixed a typo"
- the repo you "forked from" is the "original repo"
- don't say "user colon master" use the actual username
- you have a lot of filler words
- don't say "here", say "in the top of the screen" or whatever the exact location is
- don't "review this"; instead, "review this pull request"

README
- I like having the wiki page on the right!  Say it's from wikipedia.
- You don't have "how to run _your_ program" you have "how to run _my_ program"
- don't use your rshell readme example without showing how it appears on github; my recommendation is to only use your nicer rshell readme and not use the other one since it has good example information
- you should use an actual code example with the backticks
- pounds don't make things "smaller header _files_"
- you should be the first person in the contributors list

learning vim getting started
- sound is much too soft compared to music; also your screen size is messed up
- you should use `pwd` not `ls` to show that the two terminals are in the same directory
- you don't need to tell people at the beginning that vim is great; don't tell them, show them with your examples
- explain `cat`
- your examples of vim on the command line are confusing
- actually say something out loud about the keyboard on the screen; is it possible to get the key highlighting to stay on a half second or so after the key is pressed?
- don't say "pretty sure you're familiar with it"
- not "down here", but "down in the bottom left"
- there are three modes in vim: normal mode, insert mode, and visual mode.  When you say "outside of insert mode" you should say "hit escape to leave insert mode and return to normal mode"
- since the text below talks about using `G` and `gg` and `#G`, you should mention that in the video
- You should move the searching stuff in the text to another section; I think search/replace with regular expressions is one of the coolest features of vim, and I'd recommend a whole video on it.  In particular, the regexes `\<` `\>` `^` and `$` are amazingly useful.

indentation
- screen size is weird
- you should mention that you can type `5>>` in visual mode to indent 5 tabs at once that way you don't have to do it several times; at the same time mention that you can do this for any command; it'll probably be best to change the example just a little so that you actually need to indent multiple times

deletion
- your voice is very faint compared to the music
- your screen is not formatted the same way as in other videos; still have two terminals side by side even if you only use the vim terminal
- here you should remind the viewers that you can press a number before any command to have it  repeated that many times; e.g. use `2wd` to delete the whole `hello world` phrase

selecting/copy/paste
- this starts off with more movement commands, so it should be right after the first video and before indentation
- you talk about deletion here, which I think is more natural than having a separate video

appending/prepending
- I like that you're using the same text in the examples
- mention that lowercase a appends after where the cursor is, uppercase A the end of a line; lowercase i prepends before where the cursor is, uppercase I prepends at the very beginning

replacing
- I like the practical example with error messages!
- the way you said vim "prints a colon" makes it sound like vim is adding a colon to the file; always specifically say where on the screen this is happening at, e.g. "vim displays a colon in the bottom left corner and let's us type in a command"
- you specifically say uppercase-r, but you should also specifically say lowercase-r
- your terminal on the right hand side should NOT extend below the keyboard

vimrc
- again, be consistent with the screen layout; have two terminals and 
- pressing `:` does not let you add "certain stuff"; it let's you type commands
- don't be highlighting terminal lines with your mouse; vim users hate the mouse; enter visual mode to highlight things (you don't need to explain how you're doing this)
- you should show examples for the tabstop command
- don't make your example right next to the actual commands; you have a lot of space... use it!
- `{` is not a specific type of bracket called a "curly brace"; always use the more specific term when it's applicable 
- why do you automatically have `}` inserted but not `)` and `]` and so on?
- **having multiple tabs in vim is super useful and you should mention this in a video somewhere**

ls command
- I like that you have some documentation next to your terminal, but you really need a margin around the edge to make it readable
- you're saying "it" too much; be specific
- I recommend doing this within a git repo; show that `ls -a` displays a `.git` file that contains all the repo information; also use the example of the home folder; show all the hidden files that correspond to system configurations like the `.vimrc` file you just talked about.  This might be a good time to talk about the `.bash_history` file?  Tying things together like this will really help your viewers
- work through the columns from left to right; `rwx` is not a single column, but three columns
- you should have an example of a symbolic link if you're going to mention them; it would be helpful if you have a file owned by another user (e.g. root)
- the date is not when the item was first created, but when the file was last modified
- explain that there is no difference between `ls -l -a` and `ls -la` and `ls -al` and `ls -a -l`

cd/mv
- make your command list on the right bigger like you had for ls
- mention tab completion?
- **on all the videos, you have a complex prompt displayed in bash; lookup how to change the prompt and change it to something very simple (like just `$`; make sure there's the space); this will make it easier for people to follow what you're typing; also, you should be able to zoom in a bit in the console to make the text bigger; this will be useful even for people watching the high def videos**
- without any context on what the `$ cd directory/path` `$ mv filename/directory_name destination` mean, these are very confusing.  If you have them, you must have a few sentences explaining them.  I recommend just removing them.
- "If the destination argument is not a path" does not make sense; you mean if it is not an _absolute_ path

script
- you don't "press cat" you "type cat"
- this sentence is very awkward: "Have you ever wondered how you could possibly record data output from a program, or how you could record a list of commands within your terminal to share it with your colleague or professors in the future?"  Nothing else starts off with a question, so this shouldn't either

ssh
- try to make this applicable to any CS student
- you're not getting mistakes because the compiler is not the same; the compiler IS the same; they are both g++; the difference is that your mac operating system behaves differently than the linux operating system on well; that's where all your errors were coming from
- it's not "your home server", it's your home computer.  I doubt your home computer is a server
- these instructions are specific for unix-based machines and won't work on windows; you should mention that in the beginning
- the first time you ssh into a server you will get a scary warning message about encryption keys not being trusted and you have to type yes to trust them; you can delete the file `.ssh/known_hosts` (may be different on mac) to make your ssh forget that it's connected to well and ask you the question again for the video
- mention that one of the advantages of vim is that you can use it through the terminal over an ssh connection and it will still be very responsive even if you have a bad internet connection
- it might be worth talking about how to use the `scp` command to transfer files between the school server and your local machine

perror
- your `getlogin` example is pretty bad because it very rarely fails and is hard to force to fail; a better example would be `open`, which you can easily make fail by opening a file that doesn't exist without passing it the `O_CREAT` flag; then you can see how `perror` gives you useful debugging information
- that is NOT the correct way to write the size; the size passed in should be just `1024`, not `1024+1`; in face, it would be possible for an attacker to use the code you wrote to take over your whole computer; also you should ALWAYS use the `sizeof` operator (or `#define` macros) and never hardcode numbers
- use the side-by-side format so we can see you compile and see the code at the same time; use the terminal based man page in the other terminal (or have safari already open to the man page in the other side of the screen)
- never say "the way you do this"; say the "way you do proper error checking"
- you have a LOT of useless commands in your terminal that you aren't explaining and need to remove

PR #126, opened 2014-09-02
gdb tutorial added

ghost:
gdb tutorial added


mikeizbicki:
Your writing style is perfect for a tutorial like this.  But you still have a number of bugs in your writeup.  I've divided these into "small" and "big" categories.  Currently, your grade on the assignment is a 50.  Fix all the small bugs and I'll upgrade it to an 80.  There are three big bugs.  If you fix one of them, I'll give you a 90; fix two I'll give you a 105; fix all three and I'll give you a 130.

Unfortunately, you have very little time do this.  I need your resubmission **before noon tomorrow**.  If you can't make that deadline, but still want to get the grade upgrade, you'll need to talk to me so we can work something else out.

### Small
- Under "starting gdb", mention that the file already exists in the repository.  Also, the file is not properly formatted in the readme
- Technically, you don't need a program to start gdb, only to debug with it.
- You're not "making sure" `test.cpp` compiles.  You're compiling it.  You don't need to tell people to press enter.  Just say "Compile `test.cpp` with the following command:", show the command, then explain the `-g` flag.
- `(gdb)` is not a "command line".  It is a "prompt"
- This makes no sense: "The program runs normally just $ ./a.out"
- typos: Another way to run the program though gdb is to passing arguments.
- What does "already in place" mean?  I think you meant "command line parameters".
- Mention that breakpoints stop the program _before executing the line number_.
- typos: This is useful for because it suspends other lines of code from executing so that you can inspect the variables and exection of the program.
- You can also use `c` instead of `continue`.
- `breakpoint`, not `break point`
- typo; ttyping
- reword: Now there are two breakpoints in test.cpp. Now run the program from gdb by ttyping run in the command line
- what does this mean: What this does is that it pauses that this fuction, just like how it paused when we used break.
- this is not true: By running info break, there are no breakpoints showing up because we deleted them all.
- At this point: "The command list or l list ten lines of code for the current like that you are on." You show some gdb results in the middle of a program execution.  You must explain why these results are being displayed.
- This is not correct: The step or s command acts like next only if the current line is setting a variable. And if the current line is on a function, the step command will only execute the first line of the fuction.
- `finish` is particularly useful in combination with `step`, but you need to say more about why.
- reword: The step or s command acts like next only if the current line is setting a variable. And if the current line is on a function, the step command will only execute the first line of the fuction.

### Big
- Your example does not demonstrate `backtrace` very well because there are no functions in your code.  You should change the example to include a simple recursive function like fibonacci to demonstrate the stack trace.  You should also show how to set breakpoints on functions with the new code.
- You have a good discussion about breakpoints and stepping through code, but you don't explain why this is useful at all.  In particular, it lets the programmer use the `print` command to display the value of variables at different locations.  You need to provide some examples of this.
- You're tutorial misses one of the most useful aspects of gdb: finding the location of segfaults.  You should adjust your `test.cpp` file so there is actually a segfault bug caused by dereferencing a null pointer.  Then show that gdb will find the location of this bug for you automatically.

PR #130, opened 2014-09-05
Added to Lab 1 README

thekevinlemon2:
I've added some pictures for the branching examples and screenshots for the the GitHub buttons. I'm not done yet but I'm looking for feedback on my work.


mikeizbicki:
This is right on track for what I was expecting.

Content comments:
- The pictures look correct, except for one thing: They're missing the commits from the previous sections.  Start the pictures in the traveling through time section.  Then EVERY commit must be in EVERY tree you draw.
- Make sure to have a picture after the merge as well.

Style comments:
- I like how you used different backgrounds for the branch names and the commits!  But the background of the boxes should be something a bit lighter.  The purple (and even the blue) are much too dark and make it hard to read.  A dark border around a light background of the same color would probably look real nice.
- The font should match what is used on the github page: size 13px with any of the following font families: Helvetica, arial, freesans, clean, sans-serif.  Currently, all the text is being resized based ont he size of the box, but it needs to all be constant.  
- The arrows are all different sizes.  Pick one size and stick with it.
- All of the images should be centered in the text.  You can use the html `<center></center>` tags to do this.
- You need margins (padding) around the text in the boxes.  It's difficult to read letters that are right up against the side.

Github pictures:
- These pictures look very nice.  I like how you shortened some of the pictures so they only contained the relevant information.  Do that to the first image as well.  Also, make each image have only a single circle in it, and place that image directly below the text that talks about it.

thekevinlemon2:
I changed the font and colors of the pictures, but font I'm using font 15 Arial on photoshop because font 13 is kind of small. Is that okay?

thekevinlemon2:
Also I'm inserting my images via 

```
<center>! [alt text] (image_url)</center>
```

I'm not sure how to use center tags to center my pictures.

mikeizbicki:
I've updated your comment with the center tags.

Use whatever font size ends up looking the same as the regular text on github.

PR #134, opened 2014-10-03
a million typos

2c2c:
these better be worth 10 ea.


mikeizbicki:
It's customary to have only a single commit for a patch like this, and the commit message would be something like "fixed typos", not every fix you made.  Look into the `git rebase` command to combine all these commits into one and resubmit the pull request.

2c2c:
well the workflow was using a script to cat all of your readmes and compare it against a dict and output the words that seem wrong. so any context about what file i changed was lost since I did actually do them one at a time

mikeizbicki:
In another version control system, that would be correct.  But in git, the philosophy is that one commit should correspond to "one thing" and so we combine commits together whenever it makes sense.

PR #148, opened 2014-10-10
I just update README.md

mxu008:
When we update the forked repository on the github servers, I think we should run the command "git push origin cs100-2014fall" instead of "git push origin master". We should always check what branch we are in.


tkimva:
Thanks! remind Prof. Izbicki to assign you extra credit for this.

mxu008:
@mikeizbicki  I just updated README.md and sorry for the trouble.

mikeizbicki:
Good job.  Right now, you have two commits, which is fine.  In the future, try to make only a single commit for changes like this.

mikeizbicki:
+5 pts extra credit

PR #150, opened 2014-10-11
Found minor bugs in lab1-git README

KenleyArai:
1. Modified first main.cpp example. Changing iostream.h to iostream. Including iostream.h is deprecated and will not compile in g++.
2. Modified the example code for merging the branches bugfix, master, and userinput. Changing return 1; to return 0;. Not doing so would of put the code back into a bugged state.


mikeizbicki:
Nice!  We were going to talk about those mistakes in class on Tuesday.  That's awesome that you found them first.

mikeizbicki:
+8 pts extra credit

PR #155, opened 2014-10-13
Update README.md

aorti017:
Added the word "use" to make the sentence "Inside of the child process is where you’ll functions such as `exec`..." correct.


mikeizbicki:
Thanks!

PR #158, opened 2014-10-13
Small spelling edit

jesuscreyes:
Changed word from, "...it doesn't know which parts to keep from 'uesrinput'..." to "...it doesn't know which parts to keep from 'userinput'..."


mikeizbicki:
+1 pt extra credit

PR #236, opened 2014-10-14
fixed tutorial links on hw0-rshell/README.md

fvera002:
tutorial links were not working (Page not found): fixed


mikeizbicki:
I can't merge this as is because of commit aca40ef, which changes lots of completely unrelated files.  You need to remove that commit from your merge request so that only the update will be applied.

PR #242, opened 2014-10-15
Created a Script for creating an update Script

hgarc014:
The script creates an upstream to mikeizbicki's repo.
Then creates a file called update.sh, which will update the user's repo via 
git fetch and git rebase

after creating the update.sh file the createUpdateScript file deletes itself. Both files have been chmoded with 555.


2c2c:
devil's advocate: if someone uses convenience scripts made by someone else, they may never actually learn the commands that the script is abstracting away.

PR #255, opened 2014-10-18
enrolling in cs100

luizlopezm:


tkimva:
We can accept this pull request. As you missed the this week lab, you will get no credit for lab-1-git

PR #304, opened 2014-10-25
Fixed typos in various README.md files

ghost:
To Professor Izbicki:

I followed the procedures in the link that you provided regarding squashing my commits, but I am still a little confused as to why it is saying that I now have six commits, instead of the original three. Did I go wrong somewhere? Or are those just there to show that I made the respective changes?

### DIFF
--- a/assignments/lab/lab7-rm/README.md
+++ b/assignments/lab/lab7-rm/README.md
@@ -14,5 +14,5 @@ Implement your own version of the `mv` command.  Given the command:
 mv file1 file2
 ```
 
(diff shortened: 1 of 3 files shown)


mikeizbicki:
Yeah, somehow one of the steps messed up.  Commit b6cf5c2 has all the changes in it, and all the other commits are adding then removing the individual changes.  Somehow you need to delete all the commits except for that one.  The key command is `git rebase` which lets you delete the old commits, so that's what you need to google for.

PR #306, opened 2014-10-25
Update ls.cpp

aorti017:
The missing semi-colon generated two compiler time errors.


mikeizbicki:
Thanks, +5pts extra credit

PR #313, opened 2014-10-28
calc grade now installs instructor keys

hgarc014:
checkKeys iterates through all instructor files that have an @ symbol (email). all output from checking if the instructor's keys are installed is placed into a temporary file that is deleted after the loop. if the instructor key is not installed, then checkKeys calls install-instructor-keys to install the keys.

### DIFF
--- a/scripts/calcgrade.sh
+++ b/scripts/calcgrade.sh
@@ -18,6 +18,11 @@ if [ -z $user ]; then
 fi
 
 #######################################
+# check if user has installed keys
+
+checkKeys
+
+#######################################
 # calculate stats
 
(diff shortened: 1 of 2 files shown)


hgarc014:
When you say `colorPercent` should be replaced by `ColorRed`, do you want a function that will change all following output red until it is reset, or a function that would take in some text. print that text in that color and then change the terminal's color back to normal??

I assume a function that would take in text and change the color back to normal

mikeizbicki:
Either way.  It's probably more convenient to make it print whatever its parameters are in the color then reset the terminal back to normal.

PR #316, opened 2014-10-29
updated calcgrade to check for gpg

hgarc014:
Also included print color text functions.


hgarc014:
@mikeizbicki  I can create something that will allow you to have colored text in `echo` and `printf`. However, in order to get color with `echo` the `-e` flag will need to be used. Is this fine?? Also, would you like me to change some of the color usage inside?? As to adding `color` to `error` function and cleaning up other color usage?

PR #322, opened 2014-11-01
updated calcgrade and config

hgarc014:
- Cleaned up color usage.
- Added check for instructor keys
- Added check for GPG
- Added warning function
- Added color to warning and error functions

### DIFF
--- a/scripts/calcgrade.sh
+++ b/scripts/calcgrade.sh
@@ -18,6 +18,11 @@ if [ -z $user ]; then
 fi
 
 #######################################
+# check if user has installed keys
+
+checkKeys
+
+#######################################
 # calculate stats
 
 echo "finding grade for github account $user"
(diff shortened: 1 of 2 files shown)


mikeizbicki:
I've looked over your code, and am ready to merge the pull request.  Unfortunately, it conflicts with Thomas's, which I just merged.  Fix it so it can be merged with his and I'll give you 50 points extra credit.

Two other (minor) points:
1. The commit message should be more informative than just `updated calcgrade and config`.  What you wrote in the pull request is what should be in the commit.
2. Your commit is touching several different features of the code, so ideally it should be several different commits.  For example, one commit that changes only the colors, and another commit that does the gpg stuff.  This makes it easier to browse the commit log in the future.

You don't need to fix the commit issues, but if you do I'll make it 60 points extra credit instead of 50.

PR #335, opened 2014-11-05

Liniarc:


mikeizbicki:
I'm ready to merge this pull request, but I can't do it yet because it conflicts with your previous one.  Fix the conflict, then I'll merge and give you 75 points extra credit.

PR #346, opened 2014-11-05
Updated README

vdoan002:
I updated the README to reflect the new hw1 deadline and acknowledged that there is no class next Tuesday because of Veterans Day.


mikeizbicki:
You have a number of spurious commits in this pull request, so I can't merge it as is.  Otherwise it looks good.

PR #347, opened 2014-11-05
calcgrade now checks for gpg and Keys

hgarc014:
added checkKeys function to check for gpg and keys
cleaned up code for colors


mikeizbicki:
Nice!  +60 points extra credit.

PR #352, opened 2014-11-06

Liniarc:

fixed conflicts


mikeizbicki:
Commit 0c50df6 just deletes a blank space?  Remove it then I'll merge.

PR #355, opened 2014-11-06

Liniarc:

Removed key checking from uploadGrades since it will run too many times if uploading many grades at once. Run addGrader.sh to resolve signing key issues instead.

fixed conflicts


mikeizbicki:
Thanks!  +75 points extra credit.

PR #409, opened 2014-11-11
checksyscalls now checks for lstat

hgarc014:
checksyscalls was not checking for lstat and resulted in having more perrors then syscalls.


mikeizbicki:
Thanks, +3 pts extra credit.

PR #412, opened 2014-11-12
updated tests section with things from hw0 README

DustinPham:
I thought this would be nice so people wouldn't have to go back to hw0 to check what to do for testing/scripts


mikeizbicki:
It's a decent idea, but I don't want to have the same text sitting in multiple places.  That's a recipe for them getting out of date with each other.

PR #420, opened 2014-11-14
Update README.md

horventsai:


hgarc014:
It probably would have been nicer if you issued one pull request with all your changes instead of a pull request for each commit.

PR #422, opened 2014-11-14
Update README.md

horventsai:


mikeizbicki:
I'll merge these after you fix the following:
1. combine everything into a single commit and pull request
2. the name of the script should be whatever the branch name for the assignment is followed by `.script`, e.g. `ls.script` not `test-ls.script`
3. put the names of the files in backticks so they get highlighted as code

PR #432, opened 2014-11-14
Updated `pipe` section

jesuscreyes:
Removed lines of code that aren't necessary in order to use `pipe` function properly.


jesuscreyes:
Hi @mikeizbicki. I definitely want to try to complete this extra credit. Thanks for offering the opportunity.

However, I want to focus on hw#2 and get that completed to the best of my ability before I tackle this. So I would probably make the changes you suggested after I finish the assignment or by Wednesday if that's ok with you.

PR #435, opened 2014-11-15
Improvements to checksyscall.sh

anmolsingh95:
1) Checksyscalls can take in folder arguments but you have to incude path ending with /. Also its not recursive.
2) Modified the regular expressions so that open now does not match filestream member function like ifs.open("file").
3) Added sycalls ioctl, close and readlink to the list.

Bugs:The line number does not show individual file line number, but the number out of total lines. I will look into this later.

### DIFF
--- a/scripts/checksyscalls.sh
+++ b/scripts/checksyscalls.sh
@@ -1,5 +1,25 @@
 #!/bin/bash
 
+#This function takes in argument folder or files and puts them in an array called FILES. 
+#I am aware that this does not work for folders inside folders but I do not know recursion yet.
+CURR=1
+
+for f in "$@"
+do
+   if [[ -d $f ]]; then
+      for s in "$f"*
+      do
+         FILES[$CURR]=$s
+         CURR=$((CURR+1))
+      done
+ 
+   else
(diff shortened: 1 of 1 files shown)


anmolsingh95:
actually ignore this i have squashed my commits into one and will issue a new pull request

mikeizbicki:
I like the regex fixes.  It didn't even occur to me that students might be using the `stream::open` function.  What did you use it for?

Fix the following, and I'll give you 40 points extra credit:
1. Your pull request should be a single commit
2. Make checksyscalls not require the trailing `/` on directories (use the `[ -d xxx ]` command to check if a file is a directory)
2. Make checksyscalls actually go recursively into folders.  The easiest way to do this will probably be to break it up into two cases.  Case 1: if there are multiple files passed in, then call checksyscalls individually on each file.  Case 2: if only a single file is passed in, then do exactly what I already had.  This should also fix the bug about the line numbers.

anmolsingh95:
I used stream::open for the first method in cp.
2) thanks for the tip. You actually gave a very straightforward solution, while I was thinking in the wrong direction this whole time.

PR #437, opened 2014-11-15
Improvements to checksyscall.sh

anmolsingh95:
1) Checksyscalls can take in folder arguments but you have to incude path ending with /. Also its not recursive.
2) Modified the regular expressions so that open now does not match filestream member function like ifs.open("file").
3) Added sycalls ioctl, close and readlink to the list.

Bugs:The line number does not show individual file line number, but the number out of total lines. I will look into this later.


mikeizbicki:
See my notes on your other pull request.

PR #439, opened 2014-11-15
checksyscalls.sh now sucks less

anmolsingh95:
1: Checksyscalls can take in folder arguments(with or without trailing /). Also it is recursive so it will go through folders in folders.(als$


anmolsingh95:
also made some more changes, will create a pull request for that too

PR #441, opened 2014-11-15
Checksyscalls improvements and calcgrade fix

anmolsingh95:
1: Checksyscalls can take in folder arguments(with or without trailin$
2: Modified the regular expressions so that open now does not match f$
4: Renintroduced the config.sh backtrace to home
5: fixed paths to rmcomments.sh to fix the original bug
# Hope that everything works now

### DIFF
--- a/scripts/checksyscalls.sh
+++ b/scripts/checksyscalls.sh
@@ -10,21 +10,49 @@
 # by a parenthesis.  This should never overcount syscalls, but could undercount
 # them in the presence of macros, function pointers, or other fanciness.
 #
-scriptdir=`dirname "$0"`
-source "$scriptdir/config.sh"
+#scriptdir=`dirname "$0"`
+source "./config.sh"
 
-if [ -z $1 ]; then
+#Here I have described the recursive scheme as you specified.
+
+if [ $# -eq 0 ]; then
     echo "usage: $0 filetograde"
     exit 1
+
+elif [ $# -eq 1 ]; then
+	if [[ -d $1 ]]; then
+		for g in "$1"/*
+		do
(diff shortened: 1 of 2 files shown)


mikeizbicki:
This should actually be two separate commits: one for everything you'd done in the previous pull request and one for the path fix.  These are logically two separate changes you've made, and it's confusing to me to try to figure out what goes with what when I read through your patch.

I think I'm ready to merge the first commit, but it's not clear to me what you mean by the paths aren't working.  calcgrades works fine for me.

anmolsingh95:
I ran calcgrade on another computer and then by cloning a fresh copy from YOUR repo on the hammer server. It does not work due to this error
awk: cmd. line:1: fatal: cannot open file `people/students/ihund001' for reading (No such file or directory)
 ERROR: student ihund001 does not have attribute giturl in their studentinfo file
it is because it is looking for the people/students/ihund001 in the scripts directory.
Now you had a piece of code in the config file which was something like 
{until( LICENSE exists) 
    cd ..
}
This would always take you to the home directory of the repo, from where calcgrade could then find people/students/whoever.
Since you commented this code out in config, it fixed checksyscalls but created an issue in the calcgrade. Because calcgrade was designed to do stuff from the root folder.

mikeizbicki:
That's not enough info for me to reproduce the error.  You'll have to provide the exact sequence of commands you typed in.  Also, please put it in backticks so it's easier to read.

mikeizbicki:
Also, whether or not I accept your fixed paths bug, you _must_ separate it out into two separate commits.  Every commit should be about one topic, not mixed together creating two topics.  I'll up your extra credit to 50 points if I do accept both commits (because the path fix is extra from what we originally discussed).

PR #442, opened 2014-11-15
Checksyscalls improvements

anmolsingh95:
1: Checksyscalls takes in folder arguments(with or without trailing /) and goes recursively through all folders.


mikeizbicki:
Sorry, there's one last stylistic change you need to fix.  You are using tabs to indent, but everywhere else in the file uses four spaces.  Replace all your tabs with spaces and I'll merge and give you the extra credit.

anmolsingh95:
Sure.

PR #444, opened 2014-11-15
Checksyscalls improvements and calgrade fix

anmolsingh95:
1: Checksyscalls takes in folder arguments(with or without trailing /) and goes recursively through all folders.


anmolsingh95:
Hey Mike
The calcgrade fix works by reintroducing the backtrace that you commented out to fix checksyscalls. But then checksyscalls stops working again. So I specified the absolute path of rmcomments from the root folder and that fixes the error that says could not find rmcomments. 
In other words, everything works.

mikeizbicki:
Great work!  I've given you +50 pts extra credit.

anmolsingh95:
thanks.

PR #472, opened 2014-11-26
Updated Readme

vdoan002:
I tried to do this earlier in the course, but was rejected due to spurious commits.
I tried my best to revert those commits. If this is still not good enough, please let me know.


mikeizbicki:
Commit 155f821 undoes everything and redoes it, so I don't want to merge that.  Get rid of that commit then it'll be good.

PR #474, opened 2014-11-26
Added EC points possible.

gbrlrz017:


mikeizbicki:
You'll also need to merge your commits into a single commit so we maintain a clean history.

PR #496, opened 2014-12-02
added to README.md for the lab8 that you must be logged into well or els...

holidaylvr:
When I was ssh-d into hammer, bash was unable to find telnet. The same goes when I was simply on the local lab machine. I found that when I was trying to use telnet it was only accessible via the well server.


mikeizbicki:
I've just talked to systems, and telnet should be installed on hammer now.

mikeizbicki:
I'm still awarding +5 pts extra credit for the bug report.

PR #502, opened 2014-12-03
Removed a broken picture on shell-tricks

ChemicalWonders:
There was a broken picture that I found while viewing the repo. I just removed it, but I can add a different one if you would like.


mikeizbicki:
Either way is fine, but you'll need to make you're pull request into a single commit first.

PR #504, opened 2014-12-03
added getenv to checksyscall script

hgarc014:
`getenv` was not checked in the check system call script, which resulted in more perrors than system calls.


mikeizbicki:
It's pretty rare to want to be in a situation where it makes sense to call `perror` after `getenv`.

hgarc014:
true, I was just error checking it out of habbit, haha.

PR #518, opened 2014-12-05
HW4 First rough draft

mxu008:
1. This is my current status of hw4. I will add more content for if statement and loop.
2. I also find some typos in some README.md file.

### DIFF
--- a/assignments/lab/lab5-cp/README.md
+++ b/assignments/lab/lab5-cp/README.md
@@ -10,11 +10,11 @@ In this lab, you will implement a version of the `cp` command line program that 
 
 2. Use the Unix system calls `read()` and `write()` to copy the input file to the output file one character at a time.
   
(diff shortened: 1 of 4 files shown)


mikeizbicki:
1. Remove the changes you made to the labs.  If you're trying to get extra credit for them you must submit them separately.
2. The overall structure of your file is nice and well organized.
3. This section is not relevant and should be removed:  How to put shell scripts somewhere shell can find it?
4. There are NOT two types of variables.  All variables are the same.  But I like the examples you give of different variables.
5.  Your examples would be better if you picked some sort of theme.  For example, "let's write a bash script to convert the images in a folder from jpeg to png format"; then as you introduce each concept you slowly expand on the same example until you have a final finished product.  As it is, your biggest example is only 3 lines long, which isn't much of a motivation for having scripts.
6. Your writing style needs some work.  Once you get all the content done, you should get a few other people in class to proof read it for you.

PR #520, opened 2014-12-05
Added to fork, perror, and wait for hw 4

gillianzliu:
Will do at least one more later and fix any writing errors


mikeizbicki:
Your writing style is good.  And overall I think you're on the right track.

Your examples of `waitpid` and `waitid` don't demonstrate why you would want to use them ever.  They just show the basic syntax for using them.  I'm not sure how you would write a better example for these, but if you can't think of a way then you will need to more functions.

Also, since this file is getting very large, it would be nice to split it up into multiple files so that similar functions are more obviously grouped together somehow.  I'm not exactly sure what this would look like though.

PR #521, opened 2014-12-05
hw4 - contributing pseudocode

tglen002:
Contribution by:
Tiffany Glenn-Hall
Oscar Gonzalez


mikeizbicki:
1. I like the analogy to an essay outline.
2. I would provide pseudocode of some task and then c++ code of some task as a comparison.
3. Don't include the authors in the file.

PR #524, opened 2014-12-07
added vim-videos tutorial

2c2c:
I added a `vim` directory in `tutorials/` with the idea that the other vim related projects could go here as well. I went with just using markdown to link to the videos because it's probably more likely people will see the videos here.


2c2c:
pushed a dated version somehow

PR #525, opened 2014-12-07
added vim-videos tutorial

2c2c:
I added a vim directory in tutorials/ with the idea that the other vim related projects could go here as well. I went with just using markdown to link to the videos because it's probably more likely people will see the videos here.


mikeizbicki:
Nice!

The preamble is too long though.  Remove the Assumptions section and make the intro not talk about the assumptions.  Resubmit and I'll give you a 90.  The minus points are mostly because the last video is a bit disorganized.

Issue #527, opened 2014-12-07
calcgrade.sh not showing all grades


aorti017:
When I run `./scripts/calcgrade.sh aorti017` the only grade shown is for lab 0, the rest are `---`. When inspecting the grades on my ucr-cs100 repository I noticed that the cs100-2014fall branch has a grade for lab 0, but for nothing else. While the grades branch contains all the grades that it should. Could it be that the script is not checking my grades branch, but instead is checking my cs100-2014fall branch?

jesuscreyes:
I just checked my grades, and I also only have a grade for lab0.

hgarc014:
This happened to me when i first checked my grades, did you try running the script another time?

aorti017:
I have ran the script multiple times, each time producing the same result.

hgarc014:
Interesting ill try it right now

hgarc014:
Well, nothing is wrong when I run the script... If i run it for the first time it will be nothing except lab0, but If I run it again, it grabs all the correct information. So, i'm not sure what's wrong. I assume you guys are using the latest script.

anmolsingh95:
Did you try running the script from the scripts folder. For some reason, some of the scripts do not work if you are not currently in the scripts folder.

aorti017:
Running it from within the scripts folder solved the issue. Thank you.

hgarc014:
that's interesting, because I was running it within the main folder and it worked fine.

anmolsingh95:
maybe calcgrade use some form of a cache mechanism. For the first time, you have to run it from the scripts folder and then you can run it from anywhere. But take for example checksyscalls, which does not run from the main folder. That is an extra credit opportunity right there.

hgarc014:
maybe it does use some sort of cache mechanism... checksyscalls always works running from the main folder for me, haha, but it doesn't work if i run it in the scripts folder.
Well, it does work in the scripts folder if i edit the path, because the script takes you to the main folder, thus assuming your path from there...

anmolsingh95:
really, that strange. But whatever it is, a problem does exist. checksyscalls should be able to run from anywhere.

anmolsingh95:
Well. it seems that the above issue as been fixed and now all the scripts should work from anywhere on the computer. I will initiate a pull request shortly.

mikeizbicki:
The fix by @hundal78 fixes a related problem, but not this problem.

[...]

PR #537, opened 2014-12-08
All scripts run from any folder

anmolsingh95:
Fixed the problem where the scripts like calcgrade and checksyscalls could not be run from any folder except script. For more information on the problem, please see #527.


mikeizbicki:
+10 pts ec

PR #538, opened 2014-12-08
changed commit message so that grade pulling vunerability is addressed

ChemicalWonders:
included $1 at the end of the line. the commit message reads
"graded assignment using automatic scripts for $1"

hopefully this will address the issue.

### DIFF
--- a/tutorials/advanced-git/README.md
+++ b/tutorials/advanced-git/README.md
@@ -1,0 +1,152 @@
+dvanced git, and some other stuff about git
+   In this tutorial, you will learn about some small things that you should
+    be familiar with, after you finish cs100.
+    
+   Some important things that we will cover in this tutorial includes:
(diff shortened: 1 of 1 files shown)


mikeizbicki:
I'm not 100% sure that the commit messages get signed with the commits, so you'll have to point me to documentation that says so.  Also, you'll need to modify the part that verifies the signature to also verify that the username is in the commit message.

2c2c:
I figured just checking all of the commits in everyone's grades branch for unique shas would be a good approach. It's more work than this, but would ensure there's no more holes possible (besides our NSA overlords).

mikeizbicki:
@2c2c That wouldn't tell you who tried to cheat, only that someone cheated.  Also, I that could very well be vulnerable to a collision attack of some sort.

2c2c:
Alright, alright. You could always just append the name to the grade file if there's still uncertainty about it being in the commit message.

PR #541, opened 2014-12-09
calcgrade first-run bugfix

anmolsingh95:
fixes the bug where calcgrade does not show grades when run for the first time.


mikeizbicki:
I don't remember what that line was supposed to do :)  So it's possible this is introducing a bug somewhere else, but more likely I just forgot to change this line when I was doing something else.

Delete the commented lines then I'll merge.

PR #544, opened 2014-12-09
Fixes calcgrade first-run bug

anmolsingh95:
It removes the repetition of git pull that existed in my last pull request.

### DIFF
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -193,27 +193,14 @@ function downloadRepo {
     # download repo
     if [ ! -d "$clonedir" ]; then
         echo "  running git clone on [$giturl]"
-        #if [ -z "$branch" ]; then
             git clone --quiet "$giturl" "$clonedir"
-        #else
-            #git clone -b "$branch" --quiet "$giturl" "$clonedir"
-        #fi
-    else
-        echo "  running git pull in [$clonedir]"
-        cd "$clonedir"
-        git pull origin $branch --quiet > /dev/null 2> /dev/null
-        cd "$dir"
     fi
 
     # switch branch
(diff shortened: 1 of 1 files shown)


anmolsingh95:
but still its your repo so Ill just go with your style.

mikeizbicki:
I see what you're saying.  But the code below doesn't always run.  What if the `$branch` variable is empty because it wasn't passed in as an argument?

The way I have it, `git pull` can in some cases be run twice, which is inefficient.  There's probably a way to rewrite it similar to what you did that guarantees it will be run at most once but will still work when `$branch` is empty.

anmolsingh95:
Since the programmer is actually passing the branch variable, why do we have to check if its empty?

anmolsingh95:
And even before I changed anything. $BRANCH was being used in the script before checking whether its empty or not on line 204 in the config.sh file.

anmolsingh95:
Also, can you please specify what should be done in case $BRANCH is empty. That way it will be easier for me to organize things in an efficient way.

mikeizbicki:
`$branch` is not used before line 204; I don't know what you're talking about?

If 3 arguments are passed into `downloadRepo`, then the repo put itself into whatever branch is passed in as arg 3.  If only 2 args are passed in (i.e. a branch is not specified), then we stay in the default branch on clone (usually `master`, but for this repo it is `cs100-2014fall`) or whatever the previous branch was on pull.

Since the comments weren't clear to you, you should also clarify them.

anmolsingh95:
Thank you, I will add this to the comments.
`$Branch` is used on line 204 and its not been checked before.

PR #553, opened 2014-12-09
Rough draft for hw#4

jesuscreyes:
This is the rough draft for hw#4. The group members are Jesus Reyes and Itsukda Fukada.

I noticed that the .pdf file doesn't show. I was wondering if you wanted our cheatsheet to be able to display on github? If so, is there something we can do to make that happen?


mikeizbicki:
The pdf displays just fine for me.

It's a good start.  Comments:
1. You have some formatting issues with latex you need to work out.
2. Most of the things in your navigation section actually belong in your motions section.  `x` and `dd`  belong in your verb movements section.
3. Your section on folds is pretty detailed, but all of your other sections are missing a lot of commands.
4. The bit about verbs and motions should be at the top, since this is the most important, fundamental aspect of vim.  There are lots of organizational issues like this.  In general, the most important things should be at the top, and the less important things at the end.  Related things should be placed near each other.  It would probably help to draw boxes around each of your sections.

PR #555, opened 2014-12-09
calcgrade bugfix and improvements

anmolsingh95:
1) Fixes the first run bug
2) Efficient use of git clone and git pull ( runs at most once ).
3) If branch is not given, uses the default branch.

### DIFF
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -187,34 +187,31 @@ function downloadRepo {
     local clonedir="$1"
     local giturl="$2"
     local branch="$3"
-
     local dir=$(pwd)
 
     # download repo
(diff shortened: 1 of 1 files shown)


anmolsingh95:
hey @mikeizbicki the indentation looks fine on my computer. Any ideas what the problem could be?

mikeizbicki:
You should be using spaces for everything, not tabs.  Tabs will get displayed as different sizes on different computers.  If you look in my vimrc file, there's a bunch of stuff to automatically convert tabs to spaces and remove excess whitespace.

anmolsingh95:
Yes its fixed

PR #557, opened 2014-12-10
Hw4  - sed command tutorial final subsmission

asand017:
I corrected inconsistent pronoun usage and other grammatical errors in the cheating policy.


asand017:
I made a pull request this last week proposing a few corrections to the cheating policy, so that pending pull request is it lurking. Is this ok?

PR #560, opened 2014-12-10
hw4-Writing Test Cases

gbrlrz017:
So this needs a bit of work still, but here are some known bugs: 
1. My intro could use some work to give a better understanding of test cases
2. My formatting could be better. I struggled with indenting paragraphs and such.
3. I have several reminders for hyperlinks that I'll insert later throughout the file. My plan is to have a link to other students' tutorials in the places where I briefly mentioned things like debugging tools and vim. This might be hard since there won't be an actual link on/for your repo until people start submitting their final work. So I might just insert links to other tutorials I find online. 
4.  I _might_ need to shed some text. I'm thinking ~ 30% of what I have so it's a more practical read.
5. I definitely will switch out the picture I have of my table to something better and with more detail.
6. One of the more important selling points of my tutorial is the table. So I need to expand on this a bit more. 
7. Potentially Bad Jokes. 

In regards to the table, I was thinking you could allow future students to complete one as extra credit in each assignment. As a suggestion, +5 on each assignment. The points are an incentive, but they'd be ensuring their program works well while they're at it. Even without the ec, it seems like a nice worksheet to do.

What is my grade as is? And if you're willing to speculate, what is my potential grade after fixing the above things and with your suggestions? 

Thanks!


mikeizbicki:
Comments:
1.  You'll need to merge everything into a single commit for your final submission.
2. Your writing isn't organized.  Your main message is that "test cases are important" and you have a single piece of concrete advice of "fill out this tests table".  But very few of your words support this point.  So you need to get to the point much faster.  If your words are not directly addressing this topic then they shouldn't be in there.
3. You describe the tests table, but you haven't demonstrated why it's useful.  Do software developers in industry use this?  If so, give some examples of their tests on real programs and the bugs it's caught.  If not, then why not?  Do they have better methods they're using?
4. You talk a lot about why test cases are important and how to record your test cases, but you don't talk at all about how to develop good test cases.  I thought that was going to be the focus of your writeup.  Figuring out which tests to write is the hardest part about writing tests!  For example, you could take homework 0 and show all the possible ways each of the features could fail, and show how to write test cases that demonstrate that your code does not fail in those ways.  I think that would be more useful to students.
5. I can't guess what your grade will be right now other than you still need a lot of work.

gbrlrz017:
1. I agree. My main points are pretty ofuscated by the rambling. However, I feel that I did address the "how to". I'm saying that writing test cases can be split up into 3-4 categories/stages that are all important to having the program working correctly. 
   - The first stage being where you get the background on what you are supposed to implement. For example, research the system calls you need to use and run the commands you are implementing on bash to see how it handles different scenarios. These scenarios will later become the test cases, which is why I suggest that they keep track of these things on the table.
   - The second stage of test cases is when you are writing your program. The test cases that come up in this stage are usually pretty small. For example, "is this code actually parsing user input the way I want it to" or "did I just screw up the functionality of the rest of my program with the two lines I just added". There are many of these sub-test cases. Getting them right is just as important as getting the main test cases right. They are not mutually exclusive. There is no real procedure I can give them to follow in this stage as it will depend on what the reader is implementing and how their program is set up. But the hypothetical scenarios I talk about and rhetorical questions I ask are meant to get the reader thinking critically about these sub-test cases. 
   - The third category is where the reader is done writing a certain part of the program and she or he is ready to test if the main test cases for that part works. For example, testing whether the `ls -a [arg(s)]` works as it should. Here, the reader can record what happens with the test cases she or he is able to try. Having the table (or simply having it in writing) as reference lets the writer know if it is okay to move on, what works, and what needs more work. It's about organization. Again, I offer certain scenarios and some general advice to get them thinking about these cases.
   - The fourth part of my write up is where I discuss certain scenarios in which they might have to go back and forth between II and III. For example, `ls -R` and `ls -l` work fine, but `ls -l -R` has some weird behavior. So they'll have to go back and make sure sub-cases are doing what they should be doing, perhaps write some new lines of code for the interaction between these two flags, then go back to the main test case. Documenting these things seems useful because it allows you to keep track of everything. 
2. I do reference the tests table a lot without much support. Although I haven't done a good job at it, my goal is to sell the **_idea_** of the test table and documenting test cases in all stages. The tests table itself is not the greatest thing out there, but it helps the reader visualize this _idea_. This part needs a lot of work, I agree.
3. Is it okay if I talk about hw0's test cases? I was hesitant about doing the reader's work for her or him in such an example, but it would help show the usefulness of documenting test cases.
4. In #562 you mention that we can have feed back if we submit early. So if I submit tonight, will I get a grade and be able to submit again for a better grade if I request the chance to?

mikeizbicki:
My main point was that you don't have a single example.  That's what students need is what are examples of good/bad test cases.

I think you have enough to work on that you shouldn't submit tonight.  Only very minor changes will be allowed after your final submission.

gbrlrz017:
Got it.

PR #568, opened 2014-12-11
GDB Cheat Sheet Update

DustinPham:
This is what @yli082 and I have so far. If there's anything else you think we should cover, let us know!


mikeizbicki:
1. You need to combine this all into one commit.
2. There's lots of minor spelling/grammar mistakes.
3. It should be between 50%-100% longer.  You need to provide more detail
   about the commands, find other commands, or talk about common usage
   patterns/mistakes.

PR #569, opened 2014-12-11
Added Hw4, Bitwise-ops

jacksonfan:
Added for review/advice. This is nearly done, but we were hoping we could get a review just in case


jacksonfan:
Okay thanks! We'll fix this stuff right away

PR #572, opened 2014-12-11
HW4 Xmonad Tutorial Rough Draft and calgrade bug fix

anmolsingh95:
Hey mike, I have completed the rough draft for HW4. Any feedback would be appreciated.

For calgrade bugfix
1) Fixes the first run bug
2) Efficient use of git clone and git pull ( runs at most once ).
3) If branch is not given, uses the default branch.


mikeizbicki:
I like the screenshots illustrating the concepts you talk about.  That's really nice.

You have a pretty good explanation of the basics as well.

I would pick one more slightly advanced topic and discuss that.

mikeizbicki:
As for the calcgrade bugfix, I'll merge that when I merge your actual submission.  There's no easy way for me to do it separately right now.

anmolsingh95:
thanks, ill make the changes.
and also for the calcgrade, do you want to issue a seperate pull request?

anmolsingh95:
hey @mikeizbicki . Could you please suggest an advanced topic. The only one I can think of is configuring, but then I would have to teach Haskell.

2c2c:
mod , . are pretty weird

using something like dmenu as a processlauncher (doesn't come prepackaged with xmonad, but has a default key on mod p)

anmolsingh95:
Thanks @2c2c, but I have included a link in my tutorial that explains the installation and configuration of xmobar(which is similar).

2c2c:
their only similarity is that they are both bar shaped. xmobar is more for
displaying tray icons and misc info. dmenu lets you quickly launch a
program. e.g. If I want to run firefox I do `mod^p firefox enter` and
firefox runs without the baggage of having to use a command prompt first. I
don't think it's complex enough to fulfill the problem you have anyways.
mod , . deal with changing window heirarchy I think which is kind of
esoteric but seems complicated enough to fit the request

anmolsingh95:
The same step on xmobar would also run firefox. Although I agree it probably has less features than dmenu.

mikeizbicki:
1. If you submit calcgrades as a separate pull request I can merge it
   sooner, but it's fine to just wait since it's not an urgent bug.
2. I wouldn't try explaining how to write a config file.  But you could
   provide links to a number of other people's configs and explain the
   features of them.  For example, there are many different algorithms for
   tiling you can use.  You could give screenshots of what they look like
   along with a link people can follow to figure out how to set it up that way.
3. I highly recommend you email the xmonad mailing list and get some
   feedback/suggestions from them.

PR #580, opened 2014-12-11
hw4 rough draft cppcheck tutorial

khiguera:
This is our current draft. We have not finished the organization and flow of the material. We wanted to get most our material down to have a working draft. We would like some thoughts on where to improve and where to put more focus.


mikeizbicki:
You have a lot of example files, which is nice, but I'm not sure what to make of them.  They need to be the heart of your tutorial.

I highly recommend once you've made some more progress that you submit this to the cppcheck mailing list to get their feedback.  They will really appreciate that you've put in this work to explain their product, and they'll make sure everything you're saying is technically correct.

PR #586, opened 2014-12-11
squashed commits and added hw4 files

npsark:
Ok, pretty sure I successfully squashed all of my commits into one for hw4.


npsark:
I just realized I hadn't named my main tutorial README.md so I fixed that.

mikeizbicki:
Resubmit it as a single commit before I look at it.

PR #591, opened 2014-12-12
added gdb to tutorial

ghost:
Can you please look over this GDB tutorial for HW4 and provide feedback on our assignment, so we can improve the assignment and get more points out of it.

Thank you Mike


ghost:
Daniel Goerlich and Kevin Le. We worked on it together

PR #594, opened 2014-12-12
squahed commits on hlao002

hlao002:
hw4: Bash-lab by Henry Lao (hlao002) and Harry Chu (hchu007).


harrychu-git:
This is the final submission for our hw4: Bash lab.

PR #595, opened 2014-12-12
added video tutorials under vidTutorial

boxcarcoder:
Links to tutorials for system calls within this file. Submitting for feedback.

### DIFF
--- a/tutorials/syscalls/vidTutorials.md
+++ b/tutorials/syscalls/vidTutorials.md
@@ -1,0 +1,13 @@
+File Descriptors + Dup() + Dup2():
+https://www.youtube.com/watch?v=bc639jID7FU
+
+Directories:
+https://www.youtube.com/watch?v=ru3uxfknAMc
+
+Pipe():
+https://www.youtube.com/watch?v=uHH7nHkgZ4w
+
+Fork():
+https://www.youtube.com/watch?v=gese4YxnUDs
+
+


boxcarcoder:
The file vidTutorial (which contains our video links) is located in /tutorials/sycscalls. Should we have a seperate directory with our own README? I only located our links there for now for simplicity in order to get feedback.

mikeizbicki:
These videos are very well done.  Great job!  I really like the style you guys used of Paul narrating and Brendan writing.

Video 1:
- It's a little annoying how the light is reflecting on the whiteboard but I don't know how to fix something like that.
- You mention that cout is buffered and cerr is buffered; you explain the advantage of unbuffered output (it gets displayed instantly); but you don't mention the advantage of buffered output (it is faster).
- Changing the file descriptors isn't really known as "input output manipulation."
- File descriptor 0 is always called stdin, no matter whether it's coming from the terminal (keyboard), a file, or a pipe or anything else.  The correct way to phrase this is "We disconnect the terminal from stdin and connect a file to stdin to replace it."
- I wouldn't call the keyboard the "proper" stdin; it's just the default stdin.

video 2:
- everything looked good

video 3:
- awesome; I'm digging the head nod when Brendan gets introduced each time

video 4: 
- a set is a particular abstract data type, so a "set of variables" has a slightly different technical meaning than what you used (but this is minor)
- it sounds like you're saying b=0 and writing b=2?
- the child doesn't have a pid of 0, fork just returns 0 and that's how we know it's the child process; if you call `getpid` you'd get a nonzero number

If you:
1. fix those errors
2. find appropriate spots in the repo to insert your video (like the directory video in the ls assignment)
3. create a page that puts some text next to each video giving a brief overview (see https://github.com/mikeizbicki/ucr-cs100/blob/cs100-2014fall/tutorials/vim/vim-videos.md as an example)
Then I'll give you guys 120/100

PR #599, opened 2014-12-12
HW4: GDB Tutorial FIXED

ManjunathChet:
Here is the final version of my HW4 submission.
- Ex1: I changed the README so the code is seperated into different blocks. Each instruction is followed by a block of code that shows how that command is run.
- Ex2: The `info` function is now stripped of the syscalls that are used in hw2. The output is much shorter and easier to follow.
- Ex3: As you requested, example 3 is the GDB lab. I changed the introduction so it fits with the rest of the tutorial. Other than that, it is the same file.

Also, I went to the writing center to have my README's proofread. You should have the business card the tutor signed.

Please let me know if you have any other questions or concerns.

### DIFF
--- a/assignments/hw/hw4-contributing/tutorials/GDB_Tutorial/README.md
+++ b/assignments/hw/hw4-contributing/tutorials/GDB_Tutorial/README.md
@@ -1,0 +1,42 @@
+===
+
+#Introduction to GDB
+---
+
+**Description**
+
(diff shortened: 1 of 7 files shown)


mikeizbicki:
I never merged your commit because I wanted you to make those two small changes first.  Can you do that and move your files into the actual lab location on the `2015winter` branch?

PR #600, opened 2014-12-12
Hw 4 submission adding to the syscalls tutorial

gillianzliu:
I split up the original README into multiple files that group together similar system calls. I specified which system calls are in which file in the README along with a brief description of the system call. I also added another example for wait to show why one would use waitpid or waitid over just wait. This one pull request has some grammar fixes.


mikeizbicki:
looks good! 100/100

PR #602, opened 2014-12-12
Templates HW4 submission -vdoan002, amyll001, jmaw001

jmaw001:
Finished writing up templates tutorial for hw4.


mikeizbicki:
I'm not accepting this submission as is.  It contains many technical errors, and it reads like a last minute fluff piece.  I still see some of the errors I remember pointing out when I reviewed it in office hours this week.  Also, at the time, I thought this was a one person writeup and not a three person writeup.  You need to go into considerably more detail for three people.

PR #604, opened 2014-12-12
file permission for hw4

jinsongwei:
This tutorial for homework 4 is file permission. This is not final version. May change some text after meeting writing center.


mikeizbicki:
Squash your commits into a single commit before I read it.

PR #606, opened 2014-12-12
added hw4 files and squashed commits

npsark:


npsark:
Ok sounds good. I'll make the changes now.

PR #609, opened 2014-12-13
Added Hw4, Bitwise-ops

jacksonfan:
If it's possible, could @rluk001 and I have this reviewed one last time?
If not, this would be our final submission.

### DIFF
--- a/tutorials/bitwise-ops/README.md
+++ b/tutorials/bitwise-ops/README.md
@@ -1,0 +1,485 @@
+Bitwise Operator Tutorial
+===
+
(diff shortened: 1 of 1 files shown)


mikeizbicki:
Everything you say is correct, and your writing is pretty good except for the few organizational issues I point out above.  I like that you added examples to the top section, but you don't really connect those examples to the use of bitwise operations.  For example, no where do you explain that bitwise `|` is used to "add flags to a variable", and bitwise `&` is used to "check if those flags have been added."  To really make this a solid writeup, you'll have to make that connection explicit---in fact, if the reader takes only one message from your writeup, that's what it should be.  So that message needs to be incorporated everywhere.

I'm giving you both 75/100 right now, but you can fix these issues for full credit.

PR #611, opened 2014-12-13
HW#4: Updated vimcheatsheet

jesuscreyes:
This is an updated version of the vimcheatsheet from @ifuka001 and myself, ready for feedback.

Also, some fonts still need to be changed, but we wanted to get feedback on what we have so far.


mikeizbicki:
1. You should say a little more about this: "Also Certain words allow "i" to be a range which i stands for inclusion. For example,dw will delete the word from the cursor but diw will delete the entire word. "  The only reason that makes sense to me is because I already know what you're trying to say.
2. Some of your text explanations are too terse.  For example, in `:set ic -Set ignore case` it's not obvious what ignore case means.
3. Don't put the `-` between commands and their text.
4. Somehow you need to get the alignment nicer for columns whose commands are different sizes.

Other than this, your content looks fine.

jesuscreyes:
Thanks for the feedback @mikeizbicki. For the alignment issue, would it be ok to have periods separate the commands and the text, and line it up that way?

For example:

`gg` ....................................... Go to
`f+character` ......................... Finds the first instance of character

mikeizbicki:
Using `.` is very hard to do correctly.  Probably the easiest solution is to right align the commands and left align their descriptions.

jesuscreyes:
Previously when you gave us feedback, we used right align and left align commands, and you wished that the commands and their description were closer together.

Actually, I found something called a `tabular environment` that looks very promising, so I'm going to try to implement that. But if that doesn't work, I may resort to using left alignment and right alignment.

mikeizbicki:
> Previously when you gave us feedback, we used right align and left align commands, and you wished that the commands and their description were closer together.

I'm suggesting the opposite of this right now.

`tabular` is the standard way to make tables in latex, so that's what I was assuming you were using.  What were you using instead?

jesuscreyes:
Oh I see. We were not using `tabular` before. We were using the `description` environment and multiple `item` calls to list everything out.

I started implementing `tabular` and I think it looks better! It's a lot more compact, less spacing between each command, and all of the text descriptions are aligning with each other, regardless of the character-length of the command.

Update:
Actually, running into some formatting issues. I'll see what I can do.

PR #618, opened 2014-12-13
HW4 Xmonad Tutorial and calgrade bug fix

anmolsingh95:
This is the final submission of my HW4 unless there is something more you want to suggest. I have also added intructions for installing dmenu.
The pull request also has the calcgrade fix.


mikeizbicki:
The first part of your writeup (where you're explaining things with pictures) is reasonably strong, but the second half is a little shallow.  You can't provide comprehensive installation/setup instructions for setting up the network interface (for example), so you shouldn't say much about it.  Instead, show how to use one particular setup and provide links for configuring it.  Did you ever contact anyone from the xmonad community?

I'm giving you 75/100 on the writeup, but you can fix the changes for more credit.  I'm also giving you +20 on the script changes.

PR #625, opened 2014-12-14
hw4 topic about file permission

jinsongwei:
This is final hw4. Thank you professor.


mikeizbicki:
Technical content is good, but there are many English mistakes.  75/100

PR #626, opened 2014-12-14
Update README.md

Daniel-Goerlich:
Added sentence to remind students that the tag must be lower case and clarified the way to look up the functions for the extra credit


mikeizbicki:
I like 2e6ee2c, but I don't like 1c28d2f.  "Binaries" is the way most people refer to these files, so students need to understand that lingo.  Remove the latter commit then I'll merge.

PR #628, opened 2014-12-15
final submission of HW4

ogonz007:
This is the final submission for our HW4
Authors:
Tiffany Glenn-Hall (tglen002)
Oscar Gonzalez (ogonz007)

### DIFF
--- a/tutorials/pseudocode/README.md
+++ b/tutorials/pseudocode/README.md
@@ -1,0 +1,377 @@
+Why and How to Use Pseudocode
+=============================
+
+Why use pseudocode?
+-------------------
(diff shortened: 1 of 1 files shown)


mikeizbicki:
I like the 4 example types of pseudocode, but I wish there was a bit more of a logical flow between them---so that each example made it a little easier to understand the final code.  As-is, I didn't ind the pseudocode all that helpful.  

Also, there's probably a way to incorporate the biological background into the main text above in a way that flows naturally with what you're explaining, but I'm not 100% sure how to do that.

85/100, with the minus points mostly due to the pseudocode flow issue mentioned above.

PR #632, opened 2014-12-16
submit strtok tutorial

Dave864:
This is not necessarily the final submission, I am just submitting this in case my request for another extension does't go through.


mikeizbicki:
you can have the further extension

PR #636, opened 2014-12-16
HW#4 VimCheatsheet: Updated layout, fonts, description of i character, and wording for some commands

jesuscreyes:


mikeizbicki:
That second page is very chaotic.  It would be much easier to read if all the tables columns were the same size.  This will take some creativity with the formatting.

PR #650, opened 2014-12-17
final hw4 submission

khiguera:
This is the cppcheck tutorial created by @htsai010 , @lpeng003 , @afan004 , and I.


mikeizbicki:
The content is reasonable now, but the writing is a jumbled confused mess, and there are still some technical errors.  This is what your first rough draft should have looked like.  Then you go back and spend at least as much time polishing as you did writing it the first time.  65/100

PR #651, opened 2014-12-17
HW4: Fixed the beginning and follow the comments

mxu008:
Hi Mike,

I just fixed the beginning and try to follow all your comments exactly. Can your check it again? Thank you so much.

-Mengyi


mikeizbicki:
Good job.  The organization is much better now.  But there are still tons of English mistakes everywhere.  I will basically have to rewrite the thing to fix them all, and I don't think you'd learn anything from that.  I'm giving you 10 more points, but to get the rest of the points you need to fix the English.  This probably means sitting down with someone else from class and having them go through the writeup with you line by line.

PR #652, opened 2014-12-17
gnumake

nickzoo:
Thanks for the feedback.


mikeizbicki:
I already had this merged so that I could see the edits you had actually made and compare them.  You need to refactor your commit so that it includes your previous commit as well.

PR #654, opened 2014-12-17
HW4 Templates Update

jmaw001:
So we changed some parts of the README and did our best to address your suggestions. For the most part, we added a few new examples mostly under Function Templates and went more in depth for Class Templates and default parameters. If you could look this over and give us more feedback on what could be improved that would be great. Thanks!


mikeizbicki:
Let me know when you fix the changes we talked about in office hours.  Also, your final submission will need to be a single commit.

PR #658, opened 2014-12-18
changed up advanced-git tutorial README.md for hw4

ChemicalWonders:
I cleaned up the git tutorial, and tried my best at writing it as a introductory to advanced git features.

Sorry, my English grammar is not so great, but hopefully the guide makes a bit more sense than before.

### DIFF
--- a/tutorials/advanced-git/README.md
+++ b/tutorials/advanced-git/README.md
@@ -1,125 +1,163 @@
-dvanced git, and some other stuff about git
-   In this tutorial, you will learn about some small things that you should
-    be familiar with, after you finish cs100.
-    
-   Some important things that we will cover in this tutorial includes:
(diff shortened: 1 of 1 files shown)


mikeizbicki:
I'm sorry, but this doesn't seem any better to me.  I still see technical errors and the writing and organization are basically the same.  I think there are two fundamental problems:
1. You're trying to cover too much.  This makes all your explanations shallow, incorrect, and hard to follow.  The correct way to write an article like this is to start write a detailed complete explanation on one topic.  Once you finish that, move on to the next.  Don't try worry about explaining everything until you explain one thing well. 
2.  It feels like you're not trying to teach the reader anything.  Instead, it feels like you're trying to demonstrate your knowledge of git.  But that's not fun to read, and it's not going to help future cs100 students.

I'm not going to adjust your grade.

PR #667, opened 2014-12-18
Updated GDB tutorial by Kevin Le and Daniel Goerlich

ghost:
We made a considerable amount of changes from the last pull request and we would like to see what are current grade would be for this submission, along with any improvements that we can make to get a bettter score.


mikeizbicki:
This needs to be squashed into a single commit before I can look at it.

PR #668, opened 2014-12-18
hw4

ogonz007:
I think I squashed the commits correctly.


mikeizbicki:
No.  When they're squashed correctly, the pull request will show only a single commit.

PR #671, opened 2014-12-18
Removed extra files

atan009:
Only thing is the link to repo now.


mikeizbicki:
About your submission:
1. You'll need to squash the commits into a single commit.  
2. The link should be added to the lab that introduces vim with a short explanation about what it is.

About your repo
1.  "Introduction" is a really boring header.  Pretend I'm only going to read this header and nothing else.  It should still give me a flavor for what your repo does.  Something like: "A simple and effective vim configuation"
2. Explaining the difference between vi and vim is way off topic and should be removed.
3. You don't explain your vim distribution well.  What is the theme?  Why did you pick those plugins?  How is it different than other configurations?  Why should people pick yours instead of someone else's or creating their own?  I'd expect to have all those questions answered within the first two paragraphs, or I'm not going to keep reading/use your product.
4. Your plugin images are still very different from each other, and this makes the page feel ugly, unpolished, and uncared for.
5. Overall, the README is still very unattractive.  When I'm browsing the internet and land on a github page like this that lacks polish, I assume that the software also lacks polish and I probably don't want to use it.  It should only take a few hours to polish it up real nice, and then you'll have a product you can be proud to show off to future employers / students / random strangers on the internet.

PR #675, opened 2014-12-19
Fixed README.md and added template examples.

gbrlrz017:
I made the changes you suggested along with a few other minor changes. I concluded that the three images together were my best option to not distort the text size and readability, but I can try to come up with a better solution if you think it's necessary. 

Thank you!


gbrlrz017:
@mikeizbicki , you mentioned in #653 that you would give me a different score than I received if I made the changes you suggested. Was that a typo, or were my fixes not what you wanted?

mikeizbicki:
Apparently I forgot to press submit on the comments I left here earlier.  The gist is that I was giving you a 95 because: 1. the examples could still be a bit better, 2. the writing occasionally still switches tense, and 3. for the image issue.  

I've just checked your grade, and another 5 points would bump you a letter, so I've updated your grade to 100/100.

gbrlrz017:
Thank you for that!

PR #692, opened 2014-12-20
Contribution HW

atan009:
Contribution HW
Contributors: atan009, hfanc001


mikeizbicki:
I've given you both 90/100.  

1 The intro text for your repo still needs more work.  
1. Why aren't all the images in the solarized colorscheme if solorized is so good?
2.  Why'd you add your link twice to the lab file?

Otherwise, it looks nice.

Issue #694, opened 2014-12-20
HW4: Trouble squashing commits


Dave864:
Earlier I sent a pull request, #674, requesting you look at my second draft. You replied that you wouldn't look at it until I squashed the commits. I tried to do so, but in the process I ended up merging my local repository with my remote repository. Now I can't seem to try squashing again and am at a loss on what to do. I have spent about two hours trying to figure out what to do, but to no avail.

gbrlrz017:
@Dave864 I always struggled with squashing commits, especially on this assignment. These are two things that worked for me when all else failed:
1. If you happened to be working on a separate branch for this assignment, you can save the files you want to commit in a directory outside of your cloned repo. Delete the repo, then reclone the repo and `mv` or `cp` your files back into it. But be sure to do that in a new branch. 
2. In case you weren't working on a separate branch, you can still save your files outside of that repo in a folder on your computer. I personally chose to create a new repository on Github and save my files there just to make sure the format of my README.md wasn't distorted.

What I did in the second case was to reset my forked repo after I had backed up on a different repo. Run the following to do this:

git remote add upstream https://github.com/mikeizbicki/ucr-cs100.git
git fetch upstream
git branch backup ###extra insurance
git checkout upstream/cs100-2014fall -B cs100-2014fall
git push --force

Here's the reference: http://scribu.net/blog/resetting-your-github-fork.html
It's not too pretty, but it seemed to work for me. 

After you get it working, be sure to create a separate branch rooted from the original cs100-2014fall branch. Make your single commit from this new branch then run `git push origin branch_name` and you can simply create a pull request on github.

Good luck!

Dave864:
Hey, thanks for the advice. I actually used the first method to solve the issue.

PR #695, opened 2014-12-20
HW4: late strtok tutorial submission

Dave864:
In the issue I posted , #694, I stated that I was having trouble merging commits. I wasn't able to solve the problem but was able to come up with a different solution. I deleted my forked ucr-cs100 repository, re-forked it, added you and Taeyoung as collaborators again and uploaded the second draft to this new repository. I know that I am past my deadline, but may you please look at what I have?


mikeizbicki:
I'm giving you 120/100 (a little less than I promised) due to the problems with git.  But again, you've done a great job with this writeup and I'm very impressed.

PR #707, opened 2015-01-03
Changed example to exercise

ManjunathChet:
Here are the two changes you requested. Let me know if you need anything else.


mikeizbicki:
Thanks!

PR #715, opened 2015-01-08
student info yliu127

Laviness:
student info yliu127


mikeizbicki:
I can't merge this because you're deleting the exmaplestudent, which everyone else will need.

Laviness:
I've added it.

PR #717, opened 2015-01-08
Updated student info yliu127

Laviness:


Laviness:
It should be ok, github is really hard to use

PR #744, opened 2015-01-12
enrolling in cs100

ikim025:
name = Ian Hyeongil Kim
giturl = http://github.com/ikim025/ucr-cs100.git
# you MUST put valid accounts for each of the following websites below
# if you do not have an account (or do not want to list one you already have) then you may create a new, empty account

github = ikim025
stackoverflow = iankimm
slashdot = iankimm
reddit = iankimm
hackernews = iankimm


mikeizbicki:
this is incorrect; I'm not sure what you did

PR #755, opened 2015-01-12
modified the examplestudent with jinhangwang

jinhangwang:
enrolling in cs100


mikeizbicki:
the name of the file needs to be the same as the name of your cs account

PR #766, opened 2015-01-12
enrolling in cs100

Dongdongshe:
student name: Dongdong She
student ID:605276


mikeizbicki:
This isn't correct because you've deleted the `examplestudent` file.  We still need that file to be there.

PR #768, opened 2015-01-12
enrolling in cs100 and fix a spelling error

Dongdongshe:


mikeizbicki:
The repo is now in the correct state, but the history is "dirty".  You need to make this only a single commit.  At this point, the easiest way to do that is probably to reclone and make the changes again.

PR #769, opened 2015-01-12
enrolled in cs100

Laviness:


mikeizbicki:
This needs to be all in a single commit.

PR #790, opened 2015-01-14
typo fixes

tristanzickovich:


mikeizbicki:
Did you use a script to make these changes?

The vast majority of the corrections are good.  But in certain spots, you're changing the output that different programs produce.  If the program has a typo (or uses British spelling), we need to preserve that in the explanations.

Fix that and I'll merge.

PR #796, opened 2015-01-14
3 typos corrected

Fire3galaxy:


Fire3galaxy:
The second typo, you typed "marcos" at the end instead of "macros"
The third typo, you typed "esentially".

PR #819, opened 2015-01-17
Corrected and capitalized syntax under \assignments

shawnjzlee:


mikeizbicki:
Some of these changes are good.  But I personally prefer `unix` over `Unix` and `github` over `GitHub`.  They're proper nouns, but they're common enough to be uncapitalized.  So I'm not merging.

shawnjzlee:
Thank you for the feedback.

PR #830, opened 2015-01-19
fixed codeblock

Anthony1234567:
added quotations to the end of last perror argument


mikeizbicki:
Thanks for the fix!  I'm not assigning any extra credit since you've already earned 25 pts for typos.

PR #866, opened 2015-01-26
fixed grammar

mchu011:
fixed some grammar portions to make it easier to read with less confusion in sentence structure
question: is line 222 supposed to be do or do not?


mikeizbicki:
There's some good changes, but changes that need to be fixed.  Fix then I'll merge.

PR #870, opened 2015-01-26
Edited lab 3, r src/dst did not match initial command

Fire3galaxy:


mikeizbicki:
I can't merge this because you're including an old commit that I already rejected as being incorrect.

PR #872, opened 2015-01-26
Changed gdb to g++

oribix:
For some reason, the command tries to compile with gdb instead of g++


mikeizbicki:
Thanks! +1 pt extra credit

PR #883, opened 2015-01-28
Enrolling in cs100

Dongdongshe:
I thought I was enrolled in this course, but I just find that my name file didn't exist in name file.


mikeizbicki:
That's okay, but in the future, you need to check that all your pull requests have actually been accepted.

PR #887, opened 2015-01-28
Typo in bitwise-and operator

Anthony1234567:
The given value for buf.st_mode was not consistent with the value used in calculations (given: 0100644 - used: 0040644). Fixed to prevent any confusion.


mikeizbicki:
Thanks!  +1 pt extra credit

PR #891, opened 2015-01-29
Fix warnings

scohe001:
Declaring `dirName` as `char*` throws warnings that turn into errors when compiled with `-ansi -pedantic -Wall -Werror`


mikeizbicki:
Thanks for the improvement, but I actually want everyone to start with the broken code.

PR #893, opened 2015-01-29
Fix number formatting and adding tutorial

NisreenFarhoud:
Before it only counted down to 1 and kept repeating itself, now the numbers are descending.


mikeizbicki:
Fix the issues, then I'll merge.

PR #904, opened 2015-02-01
changed value in the table

ghost:
128 should be 124


mikeizbicki:
Thanks! +5 pts extra credit.

PR #934, opened 2015-02-10
Spelling error and translation addition

mchu011:
Word was spelled wrong when I looked up it's definition and I have never texted before so it took me several minutes to understand what was written.


mikeizbicki:
Thanks! +3 pts extra credit.

PR #998, opened 2015-02-19
Two file changes with error fix and add tutorial

NisreenFarhoud:


mikeizbicki:
Thanks for doing all that work for me :)

+25 pts extra credit

PR #1005, opened 2015-02-22
added cs100-runtests

M-Evans:


mikeizbicki:
This should be some relatively quick changes, then we're ready for the beta test.

PR #1013, opened 2015-02-23
code error

mchu011:
savestdin isn't declared outside of the else statement bracket


mikeizbicki:
Thanks! +1 pt extra credit

PR #1023, opened 2015-02-24
Created/Updated Table of Contents

nevarez-emil:
I made TOCs or edited them as necessary. 
I updated formatting so that it would be more consistent between files, at least as far as headers are concerned. 
I rewrote some of the files to accommodate new headers/titles. 
I rewrote the Output Formatting file to make it a bit more concise.
I cleared up some of the buggy text in the strtok file, but there is still some work to be done there.

### DIFF
--- a/textbook/assignment-help/bitwise-ops/README.md
+++ b/textbook/assignment-help/bitwise-ops/README.md
@@ -1,6 +1,26 @@
 Bitwise Operator Tutorial
 ===
 
+### Contents
+  - [Introduction](#introduction)
+  - [Practical Uses](#practical-uses)
+  - Binary Basics
+    - [Octal](#octal)
+    - [Hexadecimal](#hexadecimal)
+  - Bitwise Operators:
+    - [OR](#bitwise-or)
+    - [AND](#bitwise-and)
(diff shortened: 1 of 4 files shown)


mikeizbicki:
Thanks for the work!  This will take me a few days to go through and think about some of the edits and how we can make the style of all the files more similar.

nevarez-emil:
I kept the format pretty simple. I can put together a general guideline to my formatting style if you like, it would be about a page max. The macro/output formatting/strtok folders all share the same formatting style. Bitwise ops was the first file I changed, so it has a few formatting differences. I'll go back and fix that later though. I haven't had a chance to go through the syscalls section yet, but I'll get to it.

I was planning on going through the all the various files in the textbook folder and doing the same to them so there would be more consistency throughout this section of the repo. Many of the files already have some sort of TOC, so it wouldn't be too much work. 

Take a look at the markdown form in my repo, and let me know if there is anything else that needs to be done or changed.

mikeizbicki:
See issue #1028.  

In particular, I don't want a table of contents on each file because it makes it harder to create pdfs out of them.  Fix that then I'll merge.

PR #1033, opened 2015-02-25
spaceing issue in README

SpaceCowboy100:


mikeizbicki:
Change it to lowercase, since I use `cs100` everywhere instead of `CS100`

PR #1035, opened 2015-02-25
Prototyped Syscall Cheat Sheet

Fire3galaxy:
I have plans to add details about common errnos, structs, and flags for open(). If I missed any syscalls, my apologies.


mikeizbicki:
There are two problems with this pull request:
1. You're still including commit df05bdb, which I closed as incorrect.
2. You need to squash the rest of your commits into a single commit so I can look at them.

PR #1041, opened 2015-02-25
rough first draft

shinray:
Looks a little bare but we will add a video in the final version! any suggestions to what we should add?

todo: fix typos, clean up grammar, etc


mikeizbicki:
Overall, this is not a great start.  There's way too much fluff, very little information, and much of the information is wrong.  The first thing you need to do is come up with some good examples on using signal.  I highly recommend once you have these examples you get me to look at them again, or your final draft will not turn out good.

PR #1049, opened 2015-02-25
my first draft submission

SpaceCowboy100:


mikeizbicki:
You need to:
1. think about the outline more carefully
2. redo the examples so that future students can't just use your makefile for their assignment

PR #1054, opened 2015-02-26
Added TOC to  /assignment-help

nevarez-emil:
Also added/formatted headers within subfolders.

The TOC has interactive linking, so you can click a link and it will move you into the subfolder and down to the appropriate section.

By moving the TOC outside of the actual documents it references, you can still easily create pdfs easily and there is still a method to easily view the material and go to a specific section.

### DIFF
--- a/textbook/assignment-help/README.md
+++ b/textbook/assignment-help/README.md
@@ -1,0 +1,59 @@
+Bitwise Operator Tutorial
+---
+
+  - [Introduction](bitwise-ops/#introduction)
+  - [Practical Uses](#bitwise-ops/practical-uses)
+  - Binary Basics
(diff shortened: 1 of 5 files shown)


nevarez-emil:
I think the table of contents is a worthwhile addition because it allows the reader to preview the contents of the documents before reading through them. Instead of navigating through all of the subfolders and reading the first paragraph of the document, the reader can just simply skim the TOC and see what topics will be covered. It is most beneficial to those who are new to the folder and are finding their way around for the first time. I admit that the titles themselves are generally descriptive enough, however by adding a table of contents the reader is able to quickly access a specific portion of the document if that is all they want to see, for example if they just need to review the use of NULL in the strtok function.

I'll fix the style issues and the titles.

PR #1061, opened 2015-02-26
HW1, HW2, and lab 5 related Syscalls HW#4 Draft Retry

levinlozenges:
I worked with Lila Jomok, Github ID: goninety to produce this draft. This is a reattempt at submittng the draft.

DId not place the files in a new directory this time, just in the subfolders of the `/textbook` they were supposed to be in. Did not read that second part, sorry about that.

### DIFF
--- a/assignments/lab/lab5-cp/README.md
+++ b/assignments/lab/lab5-cp/README.md
@@ -8,15 +8,15 @@ In this lab, you will implement a version of the `cp` command line program that 
 
 1. Use the [`std::istream::get`](http://www.cplusplus.com/reference/istream/istream/get/) and [`std::ostream::put`](http://www.cplusplus.com/reference/ostream/ostream/put/) C++ functions to copy the input file to the output file one character at a time.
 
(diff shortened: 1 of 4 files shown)


mikeizbicki:
Overall, this is an excellent start.  One of the best drafts I've seen.

My recommendation is that you get away from the format of talking about each syscall individually.  It doesn't really make sense to talk about `open` without `close` and either `read` or `write`.  I think the thing will flow better if you start with an example and explain how the example works rather than go function by function.  I realize that's how the other syscall tutorials are structured, but I think that's a weakness of the other writeups, not a strength.

Likewise for the section on files/directories.  I think you should have one section on using `opendir`, `readdir`, and `closedir` all together and not function-by-function.  Then you should have another section on statting files.

If you're both motivated, I think you could finish well before the deadline and get a perfect grade.

levinlozenges:
Thanks for the feedback; I'll talk it over with Lila and we'll figure out how to make it amazing, plus fixing all those style errors. We tried to stay consistent with the structure of the past tutorials but if you're open to experimentation to the structure we can do that. I'm pretty confident that we can get this done well before the deadline.

PR #1063, opened 2015-02-27
added cs100-runtests (second time)

M-Evans:


mikeizbicki:
I think the tool is ready for a trial run.  I've posted the announcement on #1064 on how students can get the extra credit.  I'll give you two points extra credit for every student who uses the tool, up to 40 points.

PR #1067, opened 2015-02-27
HW 4 DRAFT

monti004:


mikeizbicki:
Overall your outline is good.  You are missing the most common use case of `find`, however: combining it with for loops in bash scripts.  For example, I might write:

```
for f in $(find .); do editfile $f; done
```

to run the command `editfile` on every file in the current folder recursively.  If I used the built-in bash glob like this;

```
for f in *; do editfile $f; done
```

then we'd only be running `editfile` on the files in the current folder without recursively descending into subfolders.

You have a lot of spelling and stylistic mistakes that need to be fixed.

monti004:
if i fix these mistakes can i resubmit hw4? i would resubmit hopefully by Thursday since i still have some finals to go.

PR #1070, opened 2015-02-28
Added my own .vimrc

Anthony1234567:
I would like to offer up my own vim configuration for future CS100 students to use. It has some really nice features which i think are beneficial though the repo could use some more work. It has C++ programming in mind and a really great markdown previewer that works in real time. also the colors are nice and it has a syntax checker. The entire config file can of course be customized to support other languages and filetypes.

### DIFF
--- a/assignments/lab/lab0-vim/README.md
+++ b/assignments/lab/lab0-vim/README.md
@@ -49,6 +49,8 @@ For some good ideas on what options to use, you can look at these examples:
 
 * [A vim configuration prepared by previous cs100 students](https://github.com/atan009/vimrc)
 
+* [Vimrc1234567](https://github.com/Anthony1234567/Vimrc1234567)
+
 You may also find the following links useful:
 
 * [stackoverflow: what is your most productive shortcut with vim?](http://stackoverflow.com/questions/1218390/what-is-your-most-productive-shortcut-with-vim)


mikeizbicki:
Your repo's README looks very nice with those gifs.  I like it!  How have you publicized it?

But as-is, your repo isn't very useful because you're not including a lot of information that potential users would need.  In particular, there are no installation instructions and your repo doesn't include the plugins that your `.vimrc` file depends on.  So it basically takes a vim expert to install, and a vim expert probably wants to create their own configuration and not rely on someone else.

Ultimately you need to provide a one-line do-it-all installation command at the beginning of your README file before I can accept a pull request like this.  I'd be keen to offer some extra credit for this if you streamline the installation and update the vim lab so that students are required to download and install your vim config.

Anthony1234567:
For publication, I simply created a reddit post about the repo. It has about 12 stars at the moment (all non cs100 students) 

I will update the repo so the installation is at the top.

Suggestion:
Can I create an alternate vim-lab01 where the student has the option of doing either the default lab or on their own laptop doing the lab I write. Since some of the plugins I used require the user to install separate software. I don't exactly know if the lab computers would allow them.

Or I can integrate it into the lab and remove the plugins that require the extra software.

mikeizbicki:
What extra software do you require?

Anthony1234567:
The vim-instant-markdown plugin I use requires you to node.us with npm installed then run the command:

```
[sudo] npm -g install instant-markdown-d
```

I am not sure of the lab computers would let you run the command

PR #1073, opened 2015-03-02
Adding book to ec list

NisreenFarhoud:
I think this is a very beneficial book for college students to help prepare them for job interviews.

From the back cover :
- Leads you through the job search process, including the most effective techniques for employing LinkedIn and other social networks
- Teaches you the techniques you need to solve any interview problem
- Illustrates the full thought process of application of these techniques to real interview problems with step-by-step solutions
- Solves each problem with code in C, C++, C# or Java, with emphasis on object-oriented solutions
- Includes new chapters on sorting and design patterns and new questions on 64-bit computing and secure programming practices

### DIFF
--- a/textbook/assignment-help/bitwise-ops/README.md
+++ b/textbook/assignment-help/bitwise-ops/README.md
@@ -1,5 +1,4 @@
-Bitwise Operator Tutorial
-===
+##Introduction to Bitwise Operators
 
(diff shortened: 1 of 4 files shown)


mikeizbicki:
This isn't the sort of book that you "read" though.  If you don't do the problems, you'll be done in an hour.  If you do the problems, it could take hundreds of hours.  So I like the recommendation, but I'm not sure if I'll be able to offer extra credit for it.  I'll have to think about it some more.

PR #1082, opened 2015-03-04
Adding useful info

NisreenFarhoud:


mikeizbicki:
This applies to all of the extra credit items listed, so it should be added globally not locally.

PR #1084, opened 2015-03-04
Adding useful info globally

NisreenFarhoud:


mikeizbicki:
I don't want to accept this because it's not proper grammar.

PR #1097, opened 2015-03-09
Fixed grammar

mchu011:


mikeizbicki:
The original is actually correct

PR #1106, opened 2015-03-10
Updated GDB Textbook Entry

MichaelUy:
BOUNTY: I made changes to the GDB textbook entry because of awkward writing. I made sure to keep the content of what was said while fixing the flow and clarity of the entry.


mikeizbicki:
Thanks! +7 pts ec

PR #1108, opened 2015-03-10
HW4 Update

linuxmotion:
Is this a little bit more what you were looking for. What my goal is to show is some of the advanced commands that will lead to a improved workflow, one specifically based around rebasing from the upstream repo. I of course need to finish the rebase section. Then write another section using all of these commands to simulate that workflow.


mikeizbicki:
I'll need this all squashed into a single commit for me to look at

PR #1109, opened 2015-03-10
Fixed grammar

mchu011:


mikeizbicki:
In both cases, I like the previous wording better.

PR #1110, opened 2015-03-10
Clarification and grammar

mchu011:
Fixed grammar and clarified focus of sentence


mikeizbicki:
The original text is not very good, I agree.  But I don't think your modifications really clarify it.  I'd be open to other changes though.

PR #1114, opened 2015-03-10
HW 4 Draft, Prototyped Syscall Cheat Cheat

Fire3galaxy:
Resending you my HW4 draft as I mentioned in office hours. Tried to eliminate files that were different between our branches so that they wouldn't be in this pull request. It did not turn out well, but my draft is good.


mikeizbicki:
You need to squash your commits into a single commit before I can review it.

PR #1128, opened 2015-03-11
HW4 - draft2

linuxmotion:
I wasnt sure if you wanted all the commit for this squashed. So i sent a second pull request with the all squashed. I wanted to see if this was more what you were looking for. [orignal pull request](https://github.com/mikeizbicki/ucr-cs100/pull/1108)

### DIFF
--- a/textbook/tools/git/git-tutorial.md
+++ b/textbook/tools/git/git-tutorial.md
@@ -1,0 +1,269 @@
+Welcome all, for those of you who are reading this you may have heard of a program that is 
+called git that is used to manage the source code of various projects.
+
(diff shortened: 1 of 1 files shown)


mikeizbicki:
All of the material that I saw was correct, but it needs to be more directly incorporated into the already existing git material.  My recommendation is to directly modify the git lab.  Probably the best way to do this is to insert this material between the initial git tutorial section and the section where the students make a pull request to the main repo.

Also, there are a number of spelling and formatting issues in your writeup that need fixing.

mikeizbicki:
Also, I think this is coming along well and is a lot of potential to be an amazing writeup.

PR #1141, opened 2015-03-12
Final - draft3

linuxmotion:
I made this more inline with the guidelines. I am now using the ucr-cs100 class repo as a base for the guide. I added more information to the rebase section. I think i should add a quick section about when to push to your own repo after updating your local one. I wanted to add some images about about fast forwarding, but thought it might be to much.


mikeizbicki:
Here's some options for you to think about:
- If you edit your writeup so that it's actually a part of lab1 (divide lab1 into three files; the already existing intro, then your writeup, then the already existing bit about adding yourself to the course), then you can get up to 140/100 points on this assignment.
- If you decide not to do the above, and you make all the corrections I've annotated, you can still get 100/100 on the assignment.  If you choose this option, I will not review the writeup any more before assigning a final grade.  In other words, whatever you submit next time is what I'll grade.  Your current draft is about a 50/100.  (If you choose the other option, I'll review your progress a few more times.)

Besides the mistakes I've annotated, there's lots of explanations that need some work.  I understand what you're trying to say, but someone who doesn't already know git would have no idea.  I recommend you get another student to look over it and mark down the things that don't make sense to them.

PR #1144, opened 2015-03-13
Added Boost Tokenizer tutorial

churslic:
We plan on adding a little bit more about the differences between boost/tokenizer and strtok, but otherwise we'd like some feedback on our tutorial so far.

Thanks!
Chelsey, ccali003
Virgil, vgarc018


mikeizbicki:
Adding the other examples will put you at enough content.  That shouldn't take too long.  But your writing style needs a lot of work still.  The tone of your writeup is great, but your explanations are either lacking or too verbose.  I highly recommend getting other people in class to proof read it for you.

Adding the examples and fixing the style would put you at full credit.

PR #1152, opened 2015-03-13
Final - last draft

linuxmotion:
This is the last draft pull request. I would also like to see if I can send the final pull request tomorrow. I wont be able to see any feedback until tomorrow morning. If not, everything is here, though I didn't get a chance to change the mergetool pictures to a nicer message. I believe I got all the formatting guidelines correct finally.


mikeizbicki:
I haven't had a chance to look through this very thoroughly yet, but what I've seen looks very nice so far!  Thanks!

Before I go through line-by-line, can you update the lab's overall structure so that it matches the gdb lab?  That is, the main README file is a short outline of the lab, and it provides three links inside of it to the actual lab procedures.

PR #1154, opened 2015-03-14
HW1, HW2, and lab 5 related Syscalls HW4 Submission

levinlozenges:
We split our syscall tutorials into two sections - a more traditional explanation of each system call, and an overarching demonstration to show the syscalls in action. Edited the syscall README, bitwise operators tutorial README, and the lab5 README to reflect the addition of these tutorial features. Included reference links that connect the tutorials together.

This is by no means a final submission; we'd like to hear your input on the addition of these `demos` and which format would be the most healthy and reasonable way to introduce these tutorial features. `dir.md` and `fd.md` each have their `demos` in separate files(`dir_demo.md`, `fd_demo.md`), with references to its sister files in the catergories `Follow-Up`/`Additional Resources` at the end of each file. `stat.md` encompasses both the demo and the documentation all in a single file. After critque, we'll polish up the files to use one format setting and submit that over the weekend as a final submission.

I worked with Lila Jomok (Github ID: goninety) on this submission. -Jeffrey Ng(GIthub ID: jng017)

### DIFF
--- a/assignments/lab/lab5-cp/README.md
+++ b/assignments/lab/lab5-cp/README.md
@@ -27,3 +27,11 @@ In order to determine which method is fastest, you'll need to test them on a lar
 The deliverables for this lab are more complicated than for previous labs.  We want your `cp` program added to your `rshell` project on github.  Source code should go in the `src` folder, and the `Makefile` needs to be modified to build the `cp` program and put it in the `bin` folder.  This is the same setup we had for the `ls` command you wrote for homework. 
 
 There is one catch though: you are not allowed to commit the code to your own repository.  Your partner must fork your repo, commit the code, and issue a pull request making all the changes.  I want you to get practice contributing to other people's open source projects and having other people contribute to your own projects.
+
+##additional resources
+Read the man pages on `open()`, `read()`, `write()`, and `close()` for further information on manipulating file descriptors and buffers.
+
+Here is a complete list of resources created by previous cs100 students that might help with this lab:
+
+* [Syscalls on file descriptors](../../../textbook/assignment-help/syscalls/fd.md)
+* [Bitwise Operators](../../../textbook/assignment-help/bitwise-ops/README.md)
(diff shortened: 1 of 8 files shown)


mikeizbicki:
Overall, this is really good.  If you fix all my suggestions, you can get up to 125/100 on the assignment.  In addition to the above comments:
1. I like the demos better than the other writeups.  Stick with that format and move all the information into them.
2.  Overall, your writing style lacks precision and succinctness.  That's fine for a first draft, but I want you to focus on fixing that for the final submission.  Try to get some other students to review it help you clarify some of your sentences.
3.  Your code comments in particular could use some more clarity.  It's important that they don't just repeat information that you can get from the code itself.  It should be explaining the _why_ behind things in a detailed and precise way.

levinlozenges:
Thanks for the feedback. We'll try to get another revision in soon.

I'm glad you like the demo idea, although I'm a bit confused with this bit:
`Stick with that format and move all the information into them.`
Would you like us to completely ditch the regular tutorials? Because they're quite nice in showing the libraries needed to utilize them, as well as some basic declarations of those functions(but I guess those can be found in their man pages so they're a bit redundant). Would you like them integrated in the way that it is in the `stat.md` file?

Duly noted; I have a tendency to beat around the bush in the sentences I type. I'll get some peer reviews before we send the next revision.

We went a little overboard with the comments to test the waters; we'll cut them down to size and try to leave only relevant comments for lines where the point is not clear/why we include this line of code.

mikeizbicki:
> I'm glad you like the demo idea, although I'm a bit confused with this bit:
> `Stick with that format and move all the information into them.`
> Would you like us to completely ditch the regular tutorials? Because they're quite nice in showing the libraries needed to utilize them, as well as some basic declarations of those functions(but I guess those can be found in their man pages so they're a bit redundant). Would you like them integrated in the way that it is in the stat.md file?

In the long term I'd like to transition all the syscalls pages to something more like your demo pages, so let's remove the regular pages completely.  I don't think it should be too much work for you to move their contents into the appropriate spot in the demos page.  If you think it is, then you don't have to do it.  I'd rather you focus on getting the demos perfect.

The current syscalls format is just like a watered down version of the man page.  I suspect it discourages students from using the actual manpages as a reference, which is something I don't want to do.  Ideally it should be your demo tutorials to first learn about how to use the syscalls, and then the manpages from then on out.

PR #1157, opened 2015-03-14
second draft.

shinray:
ian is working on a sigaction() video, but he's not finished.
Requested an extension to improve quality of tutorial.

### DIFF
--- a/videos/signals.md
+++ b/videos/signals.md
@@ -1,0 +1,26 @@
+UNIX-Signals
+============
+a tutorial on signals by Shinray and Ian
+
+How can vim and bash ignore `^C`? Especially considering that programs launched by bash do not ignore such signals?
+
+This page contains a few videos that should hopefully give you a better understanding of what signals are and what you can do with them.
+
+##Introduction
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=tDMjNPKQJP8" target="_blank"><img src="http://img.youtube.com/vi/tDMjNPKQJP8/0.jpg" 
+alt="IMAGE ALT TEXT HERE" width="240" height="180" border="10" /></a>
+
+In this video, we briefly mention what signals are, and give a few examples of some common signals.
+**NOTE:** `SIGSTOP` and `SIGKILL` cannot be caught or ignored.
+
+##`signal()` and custom handlers
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=GFo1Oo42Rv0" target="_blank"><img src="http://img.youtube.com/vi/GFo1Oo42Rv0/0.jpg" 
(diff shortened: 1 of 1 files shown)


mikeizbicki:
Your text section is great.  I've written my thoughts while watching your videos below.  My preference would be that you fix the comments I made in the videos below before starting on another video.  If you can fix everything in these videos, you would get full credit on the assignment.  If you are then able to make an additional video I'll give you up to 30 points extra credit, but these videos need to be fixed first.

Intro to signals:
- "method of limited communication" doesn't make sense to me
- I don't see how the drawings of a clock or the word err relate to what signals are being talked about
- You occasionally say individual letters, but I can't understand what you're saying.  You should write them out on the board as well.
- Your program should have some actual code in it... wait a second, I thought this corresponded to a program's code, but then it was erased... if this is supposed to correspond to a program's control flow instead of code, you need to make that much more clear!
- capslock values?  You had a lot of mumbled words that I couldn't hear during this transition
- it's hard to hear your voice over the music.  You should keep the music, but turn down the volume a bit.
- what does it mean that the process will interupt?
- the writing on the board is often really out of sync with the audio
- Overall the content is good.  I think it'll just take a few more practice runs to get the presentation down pat.

signal and custom handlers
- Your volume changes a lot depending on whether you're looking at the camera or the board.  I recommend you redo it as a pair again so that you can get more consistency.
- The code you write underneath `sighandler_t` isn't valid code.  Actually write out a call to `signal` and provide the error checking.
- The way you're presenting this video is by trying to show some code and explain how it works.  The whiteboard seems like an ineffective tool for doing this because you can't actually run the examples.  This video would be more effective if you either: do the presentation on a computer or come up with a more graphical explanation for using the whiteboard.  It's good to have code on the whiteboard, but you should probably have diagrams as well.
- You should add a soundtrack

kill syscall
- soundtrack is too loud.  I can't hear you
- slides don't match up to what you're talking about
- You should be in presentation mode so we don't see the red squiggly lines
- Why is it _almost_ any signal to _almost_ any process?
- When you're giving a video presentation, never say the word "this".  I don't see you gesturing, so I don't know what you're talking about.   Instead of "The kill function takes the form of this" say "the kill function's declaration is shown on the top of the screen."  The viewer needs you to specifically say the location for context
- Overall, this video feels like you're just reading the man page to me.  I could read it myself more efficiently.  What you need are examples, and you need to start with the examples and explain everything with an example.
- Your putty example is getting better.  You should make everything in this format.  But I would STRONGLY prefer it if you weren't using windows while you did this.  Also, you're wasting lots of screen real estate with blank space.  You should have two terminals open.  One with the source code, and one with the running program.  And your font size is way too small.  Fill up all the space!
- Adjust your tab size in vim to be smaller so that you get more effective screen use.

PR #1160, opened 2015-03-14
HW4:FINAL submission

Anthony1234567:
Added to textbook c++14standards and updated bitwise-ops tutorial to reflect binary information and digit separators.

### DIFF
--- a/assignments/lab/lab0-vim/README.md
+++ b/assignments/lab/lab0-vim/README.md
@@ -1,4 +1,4 @@
-# the `vim` text editor
+# The [vim](http://www.vim.org/about.php) text editor
 
 *You will need to be comfortable with the bash terminal in order to complete this lab.
 Some previous cs100 students created a [short tutorial](./terminal.md) you can work through if you need a refresher.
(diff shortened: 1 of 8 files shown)


Anthony1234567:
I added the update to the lab on vim requiring students to download and install my configuration file using  the provided one-line do-it-all installation command at the beginning of the README in my repo. The lab requires students to personalize the file themselves and are encouraged to do so in their own personal computers. I tested the command in the hammer server and commented out lines in my file that caused errors when opening up vim. The command works perfectly and does everything automatically. I took into consideration that students do not know git commands yet when they will be doing this lab so I really do walk them through the process.

mikeizbicki:
Your as-is grade would be a 65.  If you fix just the changes I marked, that would bring you to an 80.  Overall, your writing style is a bit too fluffy and hides the information that you're trying to convey.  Humor is fine, but it should make it easier to find the right material, not harder.  If you do some good editing of the writing as well, you'll get a 100.  I recommend getting other students to proof read for you.

As for the vim lab changes, you'll need to do a similar editing process to simplify the changes a lot before I give you the extra credit.

PR #1163, opened 2015-03-16
revised latex and pdf for 2 pages, finished hw 4

Fire3galaxy:
The latex file is finished and the pdf added. 

Addressing your concerns in the review: 
1. I reordered syscalls by category and relevance to class.
2. Added return values and necessary info for each syscall.
3. Revised multiline table with wordwrap and cleaner look in code and pdf. Hopefully correct format.
4. The latex file is 3 pages. The 3rd page is optional though because the syscalls are obscure.
5. I use \texttt instead of \lstlisting to produce a similar appearance and keep the latex file readable.

Thank you mike for cs100


mikeizbicki:
The material you've selected is really good!  I like that a lot!  Overall, your writing is concise and clear, although there are a handful of typos.  Fix the changes I mention above and I'd give it an 85/100.  

The only reason for the minus points is that the formatting still doesn't look as attractive to me as I think it could.  I think if you play around with it a bit more you can improve it.  If you come to office hours I can be a bit more explicit.  I also think you could fit page the page 3 functions into the two pages without removing any content.  If you can make it a bit more attractive I'll give you full credit.

PR #1164, opened 2015-03-16
Final:Updated to issue #1152, Structure changes

linuxmotion:
Added the requested structure changes. Fixed a couple verbatim typos. This is pretty much my final draft. I don't really have anymore to add unless there are some areas that you would like to have a little more on.


mikeizbicki:
Thanks! This is nice!  I'm giving you 125/140.  The reason for the 125 instead of 140 is that the images aren't consistently formatted.  They're all different sizes, and the fonts are different sizes.  Fix this formatting and I'd give you the extra 15 points.

Also, if you're trying to improve your grade, you can resubmit the piping homework during my office hours, which should improve the grade a lot.

PR #1169, opened 2015-03-17
Time Tutorial

yliu075:
I fixed the problems in `time_tutorial.md` and I organized the function examples in order of their appearance in `time_example.cpp`. I added a section on handling time in lab5 and clarified the functions. And I made the tone of the sentences more active.

### DIFF
--- a/textbook/C++/time/time_example.cpp
+++ b/textbook/C++/time/time_example.cpp
@@ -1,0 +1,57 @@
+#include <time.h>
+#include <unistd.h>
+#include <iostream>
+
+using namespace std;
+
+int main() {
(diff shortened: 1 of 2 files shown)


mikeizbicki:
Overall, it feels like you haven't addressed my main concerns, which means we probably need to speak in person about this.  Please come to my office hours. As is, this would be a 65/100.

yliu075:
Sorry I can't make it to office hours because I have gone home already. My understanding of your main concerns is that `clock()` is confusing and my tutorial does not motivate the user to use my tutorial. Can I have some feedback on my section on lab5? Should I add a section on `time_t` and `clock_t` in addition to `struct tm`?

PR #1174, opened 2015-03-18
Last Draft

tristanzickovich:
Here is my last draft, I would like my next submission to be the final one.

I took your advice from the previous draft, fixed the things you pointed out and restructured a bit.
Any other advice is appreciated!

Thanks

### DIFF
--- a/textbook/C++/boost/smartpointers.md
+++ b/textbook/C++/boost/smartpointers.md
@@ -1,0 +1,156 @@
+#Smart Pointers C++
+
+[Smart Pointers](http://www.boost.org/doc/libs/1_57_0/libs/smart_ptr/smart_ptr.htm)
+ help prevent memory errors.
+They're self-managed, automatically deleting themselves when they're no longer needed.
+In this tutorial we will focus on `scoped_ptr` and `shared_ptr`. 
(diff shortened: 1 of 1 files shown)


tristanzickovich:
Got it.  
For examples you said they should be not much longer than three lines of code, right?
Showing how to use the pointer correctly?

Would this be good code to start off my scoped pointer section:

```
   #include <boost/scoped_ptr.hpp>     //include the boost library
   boost::scoped_ptr<int> paycheckAmount(new int);     //a declared scoped pointer for a paycheck
   *paycheckAmount = 415;     //setting a value to the scoped pointer
   cout << paycheckAmount.get();     //will output the memory address of the pointer
   cout << *paycheckAmount;     //will output the value of paycheckAmount
```

Is this what it should be, more, different?  Trying to get an idea of what kind of example to show.

Thanks!

tristanzickovich:
@mikeizbicki   In case you didn't see the comment. Not sure how notifications work after a request has been closed.

mikeizbicki:
5-10 lines is okay, but definitely no longer than 10.

A great example captures both the why and the how of using the pointers.
The why aspect is usually harder to get than the how.

PR #1182, opened 2015-03-18
  HW1, HW2, and lab 5 related Syscalls HW4 Submission #3

levinlozenges:
Shortened/made dir.md and fd.md more fluid. Removed the man page signatures since they were obstructive to the fluidity of the write-up. Removed a lot of excess in sentences and left only the bare-bones of those sentences. Added links to the source code files in the write-ups.

Again, let us know if you don't like some of the formatting/some tabs were not made into spaces.

Worked with Lila Jomok (Github ID: goninety) on this submission.


mikeizbicki:
Finish the changes I've suggested and you'll get 110/100.

levinlozenges:
Thank you again for the feedback. I must have focused a bit too much on the program details part and not the "using it as an example to interactively teach syscalls" part. We'll get another commit in later tonight with those suggested changes.

PR #1186, opened 2015-03-19
Time Tutorial

yliu075:
Fixed the problems and added a section on types that includes `time_t`, `clock_t`, and `struct tm`.


mikeizbicki:
As far as I can tell, you didn't actually address any of my comments.  You still haven't motivated the examples or discussed any of the weirdness involving time like timezones or leap seconds.  I'm giving you 65/100, but you can still resubmit.

If you have questions about my comments, then please come to office hours tomorrow to discuss them.

PR #1190, opened 2015-03-19
Resize images to a almost common width

linuxmotion:
```
Images now have a nicer message
Edit image left hand placement, always on left
Update to issue #1164
```


mikeizbicki:
The images are a little better, so I'm merging.  But they still lack consistency and polish, so they're not enough to change your grade.

PR #1191, opened 2015-03-19
adding hw4

vgarc018:
Hello Mike,

We have made the changes you told us to make, and we restructured it as you said there were some structuring problems.

Thanks,

Chelsey and Virgilio
(ccali003 and vgarc018)


mikeizbicki:
This is a great final submission!  You both should be proud of your work! You get 100/100.

One last thing before I actually enter the grade though.  Update the hw0 assignment to link to the strtok writeup and to your writeup as well.  I'd like both changes to be in the same commit, which is why I'm closing your pull request right now.

PR #1200, opened 2015-03-19
What Do You Think?

admiralakbar:
I've added Info on Functions, POSFIX, and some examples of cool Bash stuff. Let me know of anything else that I should include, elaborate on, or change? I've tried to avoid messing with the original material, as I don't know if its okay to change, but if you want I can go in and change some style errors and expand upon other things, he did not discuss.


mikeizbicki:
You need to squash all these commits for me to review it.

PR #1203, opened 2015-03-19
New branch with new files. Old branch got messed up in rebases....

Fire3galaxy:
Made edits to make the pdf more readable and follow your idea of associating wait with fork and exec. Using partial lines to still delineate between different kinds of functions and make reading function descriptions. None of the words should be in the lines anymore either.


Fire3galaxy:
If you are now grading it, give me 5 minutes to commit something. I've added something fun!

mikeizbicki:
There's still lots of inconsistencies in the formatting.  I also really don't like those half-way across lines.

Fire3galaxy:
Mike, can you help me understand what formatting concerns you have? I'm pretty inexperienced with latex and also don't see the issues you have. I thought that the half-way across lines would help distinguish wait, exec, and fork without separating them with a full-way horizontal line. A dashed line would look strange and no lines makes it a little text heavy to me. 

Also for the brawl icons, thanks for looking at that. I put the tex file in the zip because it comes with png files needed to compile the latex file. I thought that having it in a zip file was more preferable than committing 8 brawl icons in github. Also, the brawl icons were intended to give the cs100 students a bit of levity in enjoying the cheatsheet while distinguishing the major categories from each other. They could replace horizontal lines, but I guess that isn't a great idea. Is the levity ok, or is it too unrelated?

EDIT: I readded the vertical line because then the half-way horizontal line doesn't look incomplete. It also accomplishes really making the cheatsheet more readable and pretty. Please help me find the rest of the formatting concerns.

mikeizbicki:
> Mike, can you help me understand what formatting concerns you have? I'm pretty inexperienced with latex and also don't see the issues you have. I thought that the half-way across lines would help distinguish wait, exec, and fork without separating them with a full-way horizontal line. A dashed line would look strange and no lines makes it a little text heavy to me. 

I understand what you were trying to do with it, and I agree that dashed lines would probably not look very nice.  But I don't think the half way lines look very nice either.  I think you need to bold some of the lines, or double the lines, or something else.

> Also for the brawl icons, thanks for looking at that. I put the tex file in the zip because it comes with png files needed to compile the latex file. I thought that having it in a zip file was more preferable than committing 8 brawl icons in github. Also, the brawl icons were intended to give the cs100 students a bit of levity in enjoying the cheatsheet while distinguishing the major categories from each other. They could replace horizontal lines, but I guess that isn't a great idea. Is the levity ok, or is it too unrelated?

Putting things in a zip file ruins git's ability to track diffs and for people to easily add changes.  As for the icons, Levity is good, but it needs to somehow be related, otherwise it's just distracting.

Fire3galaxy:
Hey Mike, would it be ok to show you possible pdfs more often? I'd appreciate the response.

PR #1210, opened 2015-03-20
Hw4 map tutorial

oribix:
STL map tutorial featuring function pointers


oribix:
on the word "implement", should i take out the word or elaborate on it?

PR #1218, opened 2015-03-20
HW 4: Vertical Lines, bolder borders

Fire3galaxy:
Hey Mike, hope you like the new edits. The bold borders help make the document more appealing to look at and make the lines on the inside easier to see. The partial horizontal lines are kept because, combined with the vertical line, they help distinguish between functions without looking ugly or incomplete. I think this really works. The vertical lines also really create a clear box in which related functions in the left column are clearly together.


Fire3galaxy:
Let me squash commits first.

PR #1219, opened 2015-03-20
Added bolder borders and vertical line

Fire3galaxy:
Hey Mike, I created a new branch instead because I didn't know how to rebase with binary files. Even rediting the latex file seemed cumbersome compared to creating a new branch.
-    I've added thickness to the table's borders. This makes the lines inside the table much easier to see.
-    I've also reinserted a vertical line between columns 1 and 2 to justify a partial horizontala line. The functions in the left column are clearly separated by group and the descriptions in the right column are separated and easier to read by individual function.


mikeizbicki:
It looks nicer with the vertical bar, but the layout is still confusing to me, so I can't upgrade your grade.  Also, I just noticed there's an error in your description of relative and absolute paths.  An absolute path is a path that begins with a `/`, and a relative path is any path not beginning with `/`.  All your examples are relative paths.

I'm giving you 90/100.  I won't have time to look at any further changes you make.

Fire3galaxy:
Thank you Mike. I can make a quick fix and resubmit if you'd be ok with that.

mikeizbicki:
There's no limit to the resubmitting :)

mikeizbicki:
I always like getting better versions :)

PR #1220, opened 2015-03-20
third draft

shinray:
Made suggested changes (except for sigaction video). Ian is working on that

### DIFF
--- a/videos/signals.md
+++ b/videos/signals.md
@@ -1,0 +1,31 @@
+UNIX-Signals
+============
+
+How can vim and bash ignore `^C`? Especially considering that programs launched by bash do not ignore such signals?
+
+This page contains a few videos that should hopefully give you a better understanding of what signals are and what you can do with them.
+
+##Introduction
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=LgNg_fyk8-c" target="_blank"><img src="http://img.youtube.com/vi/LgNg_fyk8-c/0.jpg" 
+alt="IMAGE ALT TEXT HERE" width="240" height="180" border="10" /></a>
+
+In this video, we briefly mention what signals are, and give a few examples of some common signals.
+**NOTE:** `SIGSTOP` and `SIGKILL` cannot be caught or ignored.
+
+##`signal()` and custom handlers
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=mzdCf0zCmrg" target="_blank"><img src="http://img.youtube.com/vi/mzdCf0zCmrg/0.jpg" 
+alt="IMAGE ALT TEXT HERE" width="240" height="180" border="10" /></a>
+
+Here, we demonstrate how to catch a signal, and also how to override them with our own custom handler.
+
+##`sigaction()` and an example
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=ahRBRGVTi5w" target="_blank"><img src="http://img.youtube.com/vi/ahRBRGVTi5w/0.jpg" 
+alt="IMAGE ALT TEXT HERE" width="240" height="180" border="10" /></a>
+
+This is a simple lesson on what sigaction is and how sigaction can be used.
+##`kill()`
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=O3UZqUTB5-w" target="_blank"><img src="http://img.youtube.com/vi/O3UZqUTB5-w/0.jpg" 
+alt="IMAGE ALT TEXT HERE" width="240" height="180" border="10" /></a>
+
+Here is a lesson on the `kill()` function. `kill()` is used to send a specified signal to a specified process.
+[Code used in the video](http://pastebin.com/dq546zkG)


mikeizbicki:
Overall, the material is great, and so is the presentation.  Your voices are clear and well paced.  The volume is maybe a bit too loud now, but that's not a huge issue.  It's much easier to turn the volume down.  With that said, I have a number of comments.

signals 
- You use the words "control signals" to define what a signal is, giving a circular definition
- Don't mention atomicity; that's much too advanced for an intro to signals
- Your code has giant black borders on every side.  Get rid of these borders and make your screen larger.
- Never say the words "here".  In your final submission, I will give a minus 20% everytime I hear this word.  Instead, say "In the right window we have an example program".  Give the viewer some context.
- You mention SIGTSTP, SIGCONT, and SIGINT but don't explain anything about them.  This is very bad.  Don't mention these things until the absolute last second when you need them.  First describe what the program does.
- In the terminal on the left, your prompt takes up over half the line!!!!  Shorten your prompt to just `$` like I did in class.
- You need to have the letters `SIGTSTP` actually written out somewhere for the reader to see visually.  One way to do this is to just type it in the terminal while you're saying it.
- Your slide page with the signals is great.  The only comment I have is that you haven't introduced what the `bg` command does.  You should either not mention it or figure out a way to demo it.  Also, you say `SIGSTOP` cannot be caught or ignored, but you haven't yet explained what it means to catch a signal.

singal()
- A system call us not "used for error checking"
- You're doing a better job of gesturing with the mouse to attract my attention to what you're talking about.  But your mouse is relatively small compared to the text, so it's a little hard to see.  Maybe if you actually highlight the text it would be more obvious.
- I've never heard the phrase "signal disposition" before.  Did you find that in the documentation somewhere?
- Don't mention SIG_IGN or SIG_DFL at the very beginning.  First finish explaining your basic example, then show the more complex cases that use these options.  In general, it seems like you have a tendency to get into complex details a little too soon.  I like that you're being thorough, but you just need to rearrange the material a little.
- It does not "block" your original signal; the correct phrase is "When the signal is _raised_, the operating system will call your signal handler function."
- It's a annoying that I can't see all the code in the right panel.  If you get rid of the borders, then I should be able to see it all with the font staying the same size.
- I would like to see this video merged with the previous one.  I think the content is related enough that it will make it low a little better.  Also, other videos people have made use the convention that the code is on the left and the terminal is on the right, so follow the same convention.

sigaction:
- Never say 'in many ways'; just say how it's similar, i.e. they both register signal handlers
- You need to say somewhere in the intro that `signal` is not recommended because it behaves differently on every operating system.  `sigaction` is a little more complex, but it's guaranteed to work exactly as you'd expect when you change operating systems.
- Put "perrors on the errors" doesn't make sense
- Never say "structure" always say "struct"; Never list the elements of a struct as a bullet list.  It's much easier for a programmer to look at the actual source code.  You can put comments next to each internal variable to describe what they do.
- You need to put all your code in a fixed width font so it's easy to tell that it's code
- Your example program needs to be _exactly_ the same as the example program from the previous signal videos.  You also need to use the same screen layout of code on the left and a terminal on the right.  You're wasting a lot of space as-is.
- Never say "here".  See above.
- You didn't use any of the variables within the struct.  So why did you explain them?  You need to modify the example to use them somehow.
- There's too much power point at the beginning.  You need to start with a 15 second code demo to give the viewer some context.  Then switch to explaining the power point. Then go back to explaining the code.  This is also true for the first video above.

kill
- You need to modify your vim settings so that it's the same as the above two videos.  Again, get rid of the borders so I can actually see what's going on.
- "The function will look something like this" doesn't make sense.
- Don't talk about special cases before showing the example for the most basic case!
- Never say the word "this".
- "Permissions allowing of course"  You can't just throw something like this out there.  You would have to give a detailed explanation of when permissions would and wouldn't allow this, and how you change that.  Likewise, you can't use the phrase "everyone in the caller's group" without a detailed explanation about what this means.
- Again, don't talk about sig 0 until after you've explained the examples.
- "Error checking is still performed even if sig is 0"  This makes no sense.
- Your example code isn't very strong.  I can't think of a good reason to call `kill` on your own process.  Why not just use `exit`?  You'll have to actually fork to create a child, and use `kill` to send signals from the parent to the child.
- Executables never have the `.o` file extension; this is for object files, which are the files created by g++ before linking occurs.

grades
- The amount of material is less than I hoped for from a group of 3.  If you fix all the above changes, the best possible grade I can give you is an 80/100.  If you want a higher grade, you'll need to add some more content.  Some ideas are: sending signals from bash (i.e. using bash's kill command) or talking about different signals (SIGALRM, SIGCHLD, SIGHUP, SIGPIPE, and SIGUSR1 would all be good choices).  You wouldn't have to do an additional video each; I'd assume that you're all helping each other out behind the scenes.  If you do decide to do this, I recommend sending me a rough outline first so that I can tell you how what's the maximum amount of points you can expect.  Extra credit is certainly possible.

shinray:
Ok so when I say "here", I actually point with my mouse and I try to highlight the text so it's clearly visible. Is that not enough?

shinray:
Also, this was not a group of 3. I don't know where you got this from but there is only me and Ian

mikeizbicki:
It is not enough to point with the mouse and say "here"; it takes no extra time to say "the right panel" and that provides significantly more context.  You need to always give both audible and visual cues.

Sorry, I just typoed 3 instead of 2.  Everything else I wrote in the paragraph was assuming you were working as a group of 2.

shinray:
about the signal() "blocked", I was going on information from http://linux.die.net/man/2/signal which says "If the disposition is set to a function, then first either the disposition is reset to SIG_DFL, or the signal is blocked (see Portability below), and then handler is called with argument signum. If invocation of the handler caused the signal to be blocked, then the signal is unblocked upon return from the handler." Is my interpretation incorrect?

How long do we have to finish this? It's already Friday, and I'm not sure if I can do any more voice overs as I'm starting to get a little sick

shinray:
Having to rewrite a script, record a video, edit the video, and render and upload to youtube for feedback is somewhat inefficient, is there any way to get feedback on our information that's faster?I could mock up a script for the content.

mikeizbicki:
I'm not what your "interpretation" question means.  You have a big quote,
but I don't see where you're interpreting the quote.

I'm submitting grades on Sunday, so that's the absolute last chance.

I can review your scripts for you if you want.

shinray:
> It does not "block" your original signal; the correct phrase is "When the signal is raised, the operating  system will call your signal handler function."

On linux.die.net, the man page for signal mentions 

> If the disposition is set to a function, then first either the disposition is reset to SIG_DFL, or the signal  is blocked (see Portability below), and then handler is called with argument signum.

So I assume that means that when signal() sets a disposition to a custom handler, the original signal is blocked temporarily and the handler is executed in its stead

mikeizbicki:
Ah, I see.  This is actually a really good question.

"Blocking" is something that happens to the process's ability to receive
signals, not to individual signal handlers.  The signal is blocked means
that the signal cannot be sent to the process.  So, for example, if your
process receives a SIGINT, some operating systems might block your process
from receiving another SIGINT signal until your first signal handler
returns.

PR #1227, opened 2015-03-20
HW4 and added ec reading links

kavj:
Both were merged in my commits, and I'm not sure how to exclude the reading list update.


mikeizbicki:
This will all need to be a single commit before I can accept it.  Also, I'm reposting my email here: 

 I don't think the tail, head, and split programs
will be deep enough for you to write something meaningful.  I thought
we had agreed you would use the programs categorized as "modified
program invocation":

chroot invocation: Run a command with a different root directory
env invocation: Run a command in a modified environment
nice invocation: Run a command with modified niceness
nohup invocation: Run a command immune to hangups
stdbuf invocation: Run a command with modified I/O buffering
timeout invocation: Run a command with a time limit

Is there a reason you chose not to do these programs?  All you need to
do is come up with one example for each program.

PR #1229, opened 2015-03-21
HW4 map tutorial

oribix:
I condensed the guide more. I ran a grammar checker on the guide . The guide now also comes with two example .cpp files. The first one is the ls example from the guide. The second .cpp file is an example of a map used to output month numbers.

I will make more example .cpp files. this pull request is to give you a status update before  commit too much time to the .cpp files with an unsatisfactory guide.

I have noticed that specific words seem to tip my guide in the wrong direction. For example, in my last pull request, I used the word "implement", and it  confused me. I didn't know if my guide was lacking in information or if "implement" was the wrong word to use. If a situation like that arises again, it would help if you elaborated on what I should do.

In regards to the "implement" comment you made on my last pull request, I changed the word to "use".   Do I go back to "implement" and try to make my guide teach how to "implement" or do I change the sentence to better reflect my current guide?


mikeizbicki:
The writeup you have is starting to shape up pretty nice.  You're doing a good job taking the feedback.

I said "implement" was a bad word because you're not implementing a map.  Implementing means providing the code that makes it work.  The stl library implemented the map, and you're just using it.  Using maps is more appropriate for this class.

As for the cpp files: If you're not going to discuss them in the writeup, then don't work on them.  I won't really look at them.  If you want to add additional content, it should be in the README itself.

If you fix the minor corrections above, then you're at a 75/100.  In order to improve the grade, you'll have to add more content.  In particular, I'd like to see more clever uses of maps.

PR #1232, opened 2015-03-21
hw4 submit

SpaceCowboy100:
I've made the changes suggested in my previous pull request https://github.com/mikeizbicki/ucr-cs100/pull/1207


mikeizbicki:
The writeup is much better than before, but I still don't think it improves the existing writeup.  It has a number of technical and stylistic errors as I've annotated above.  I'm giving you a 65/100 for now, but you can still resubmit for more credit.

PR #1241, opened 2015-03-21
Hw4 map tutorial

oribix:
I changed the things as you requested. I couldn't find the function pointer write up that you requested, so I linked to the tutorial that I used to learn about function pointers (I didn't even know they existed until you asked me to include them. Seriously, I am getting bang for my buck in this class). Currently I am also looking into other uses for maps.

One use that I thought of is a (really bad) login system. Is there any merit to this idea or should I trash it? I'll keep looking.


mikeizbicki:
I'm glad you're learning a lot :)

Can you also update the ls assignment to include a link to your writeup?

PR #1249, opened 2015-03-21
Hw4

GitHubbubble:
Pretty proud of it myself actually, wish I'd gotten more data with how cool it looks


mikeizbicki:
This looks really nice!  Great job!  A few comments:
- I'd like this added directly to gitlearn.  In that repo there's a folder
  called `scripts`.  Don't put it there.  Create a new folder called snap and
  put it all under there.  It will need to be in a single commit.
- You have your axes backwards.  The grades should be the y axis and
  everything else the x axis.
- I think you're missing a number of images from your readme.  The text
  makes it look like there's supposed to be something there, but it doesn't
  load.
- It will take me a while to look through all the specific graphs, but I'll
  probably have more to say after that.

I'd like to keep working on this with you if you're interested.  I have a
number of ideas for how we can keep adding to this.

GitHubbubble:
Mike I am so interested, wow. I would love to keep working, I already have a lot of ideas! I Thank you so much for the feedback on my new nocturnal schedule!! I'm gonna whip up these changes right now! Wow. 

It'll take me a little time to get graphs in the right order, I have to transfer them. Shouldn't be more than an hour but

PR #1254, opened 2015-03-21
Hw4 map tutorial

oribix:
I changed the things that you told me to. I am still looking around for another cool way to use maps.


oribix:
also I found a stray semicolon in the HW1 Readme

PR #1256, opened 2015-03-21
Fixed mistake in description of absolute and relative paths. Removed par...

Fire3galaxy:
...tial horizontal lines.

I realized that the partial horizontal lines weren't actually self-explanatory. Guess I really thought they were helpful. Removed them for now. I think the boxes are readable enough for future students. Thanks Mike.


mikeizbicki:
I can't merge this commit as-is because it doesn't build off your previous commit.  This prevents github's automerge feature.

PR #1263, opened 2015-03-22
My updated gdb Tutorial

ahvino:
I am hoping I can get the ok to do the fork() video as well. In the meantime here is a polished version


mikeizbicki:
I agree that these videos are much improved.  I'm upgrading your grade to 110/100.  The reason it's not 120 is that there are still a few minor technical and stylistic errors.

PR #1279, opened 2015-03-27
hw4 for review and extra credit reading

kavj:
I'm trying to figure out better examples. I've reviewed a lot of literature in the past day. Is there any feedback you can suggest for something more practical?


kavj:
oops one file came in twice. I'll fix that.

PR #1294, opened 2015-04-06
enrolling in cs100

kchan049:


kchan049:
added "fixed the spelling of Linus Torvalds" commit

PR #1403, opened 2015-04-17
fixed confusing wording in README

laurenB93:
I did not mean to add the lab2 file.


mikeizbicki:
You need to recreate the commit so it only contains the material that should get updated.

PR #1417, opened 2015-04-18
valgrind isn't being tested in hw0

kluu006:


mikeizbicki:
Thanks for the edit.  But since this note is only in the comments, I want to keep it in for future classes.

PR #1430, opened 2015-04-20
fixed a nonworking link

kevdev158:
fixed '#additional resources' to '#additional-resources' which redirects the user to the right place in the page instead of nowhere.


mikeizbicki:
Thanks! +1 pt ec.

PR #1433, opened 2015-04-21
added delete and info break

markasfour:
Added information on how to delete a breakpoint and how to view the status of all of your breakpoints.


mikeizbicki:
This is great info, but it should come later in the lab.  (It doesn't make sense to talk about deleting breakpoints before we've even seen an example of how to use them!)  Also, both of these commands should have examples.  Make these changes and I'll give you +15 points extra credit.

PR #1437, opened 2015-04-22
it should say hello not goodbye

divyanshch:
The output when running the program should be 

```
First: hello
Second : hello
```

not 

```
First: goodbye
Second : goodbye
```


mikeizbicki:
Thanks!  +5 pts ec.

PR #1451, opened 2015-04-27
Updated mchen046 student file

mchen046:


mikeizbicki:
Thanks.  Your grades should work now.

PR #1500, opened 2015-05-06
fixed download GDB link

amah001:
GDB link doesn't work and leads to a github 404 error.
I added "http://" which fixed the link


mikeizbicki:
Thanks! +1 pt ec.

PR #1535, opened 2015-05-12
Added closing statement in parent 2

kevdev158:
Need to close write end of pipe in second parent so the child of that process is not left waiting for input from write end of pipe. I was playing around with the code on here and I was wondering for the longest time why after my second `execvp()` the program was dangling.

I tested using: `ls -al | cat` which should have outputted whatever ls -al did, which it did, but it also left the program constantly asking for input.

Source: http://stackoverflow.com/questions/5610184/input-redirection-problem-while-using-execvp


mikeizbicki:
Great catch! +10 pts ec.

PR #1562, opened 2015-05-17
Added missing comma in git blame line range format

edyionescu:
`<start>`,`<end>` line numbers must be comma-separated.


mikeizbicki:
Thanks!

PR #1567, opened 2015-05-18
revised regex tutorial

htang014:
I personally found the tutorial to be very hard to follow.  I have edited bits and pieces but please look over them


mikeizbicki:
Thanks! +5 pts ec.

PR #1575, opened 2015-05-20
typos? in sed

cdixi001:
some things that annoyed me. these are the first typos i am submitting. I finally learned how to properly submit pull requests after a few labs...


mikeizbicki:
you'll need to squash all these commits into a single commit before I can look at this

PR #1577, opened 2015-05-20
typos

cdixi001:
i'm not sure if i'm doing the commits right. i think i fixed all the things you asked for in your comments. I had a lot of trouble with the squashing, so if this isn't right, I'll have to figure something else out..


mikeizbicki:
We want to keep our git histories as "clean" as possible.  This means that we don't want to add commits that are going to immediately be undone by other commits we're adding.

You'll need to actually remove your previous commits before creating new commits.  There are a number of ways to do this.  One simple way is to use `git reset` to rollback your history to before you created the unwanted commits.

PR #1578, opened 2015-05-20
typos (squashed)

cdixi001:
i think i squashed the typos finally. thanks for the help, it took me some time and frustration, but i did learn a lot of new stuff


mikeizbicki:
Thanks!  +9 pts ec. 

In the future, you would also want to get rid of the merge commit 77c5f62 as well.  Since it's there, I can't use the automatic merging feature.  I have to run `git cherry-pick` and merge commits manually.  This is good for your first pull request though.

PR #1580, opened 2015-05-20
Added "correct" example of a quiz question

DoranCals:
Here's an example of a student's fixed code. I'll introduce a few commonly-seen syntax and logic errors into this file, and then write another "program" full of functional but unreadable code for students to decipher.

### DIFF
--- a/textbook/splask/README.md
+++ b/textbook/splask/README.md
@@ -1,0 +1,13 @@
+# splask
+## C++ logic and style quiz
+
+The "SeePlusplus Logic And Style Kwiz" is a set of general-purpose programming exercises designed for beginning and intermediate programmers.
+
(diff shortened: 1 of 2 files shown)


mikeizbicki:
I'm sorry, but this is going in the wrong direction.  The problems you have are very different than the quiz we had in class.  In particular, your quizes are very basic.  They are about at the cs10 level.

The code from the in class quiz did not ask you to find syntax mistakes or basic style mistakes like poor variable names.  Instead, everything was done somewhat reasonably.  But the code exposed some of the weird edge cases of the syscalls, and didn't handle those weird edge cases correctly.  Your quizes are supposed to be at this higher level.

Does that make sense?

PR #1591, opened 2015-05-20
gdb and strace rough drafts

amah001:
gdb and strace outlines/scripts for the videos along with predicted changes are in comments


amah001:
Did you check our pull request yet? Also, did we put our submission in the correct spot?

mikeizbicki:
This is the right spot.  I just haven't had a chance to look at anything involving videos yet.  That should happen late tonight.

PR #1592, opened 2015-05-20
HW4 Rough Draft :]

jle034:
Hi Mike. 
This is what I have written up for my grep tutorial so far. I was having trouble with wording and organizing it, so let me know how it looks so far. I also couldn't squash my commits because i wanted to push and see how the markdown formatting looked. 
I am still going to add to the introduction, and i'm going to include information about: 
- commonly used flags
- features of grep (using '.', '$', '^', etc.)
- when to use grep (not sure if I should include something like this)

### DIFF
--- a/textbook/tools/bash/grep/README.md
+++ b/textbook/tools/bash/grep/README.md
@@ -1,0 +1,49 @@
+##Grep
+
+####What is grep?
+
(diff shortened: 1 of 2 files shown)


mikeizbicki:
The examples you select are the most important part of the writeup.  Once you have your examples, let me know so I can take a look at them.

PR #1594, opened 2015-05-20
Added gdb cheatsheet

donei003:


mikeizbicki:
Sorry I didn't reply to this earlier.  Your idea seems fine, but I recommend you show me another draft before the final deadline.

PR #1595, opened 2015-05-20
Rough Draft HW4, began integration

tortas:
Very much a work in progress. I began compiling the two together into one document. Now I need to go through and edit them for technical accuracy and style. Then, I will add any features that weren't included that I feel would be a useful addition.


mikeizbicki:
Seems like a reasonable start.  Let me know if you have any questions about it.

PR #1597, opened 2015-05-20
First draft of HW4, Boost Tokenizer tutorialk

glaceeeee:
Wrote escaped list and offset tutorials and added to the README file

To do:
- Add examples to escaped list and offset tutorials
- Edit the README for better flow
- Add parts on the constructors for the escaped list and offset tutorials
- Cut down on some wordiness/ get to the point faster (mainly for the escaped list tutorial)

### DIFF
--- a/textbook/assignment-help/BoostTokenizer/README.md
+++ b/textbook/assignment-help/BoostTokenizer/README.md
@@ -1,9 +1,56 @@
 #The Boost Library Tokenizer
 
+>"...one of the most highly regarded and expertly designed C++ library projects in the world"
+> - Herb Sutter and Andrei Alexandrescu, C++ Coding Standards
+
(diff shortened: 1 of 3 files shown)


glaceeeee:
Also, I think I'm going to take out the quote in the beginning of the README.

mikeizbicki:
The final product should just be a single file, not one for each class.

mikeizbicki:
The main thing you need to focus on is incorporating your material more seamlessly into the existing tutorial.  It shouldn't feel like something that was tacked on afterward.

Any questions?

glaceeeee:
Thanks for the feedback!

Are you going to be available to read later drafts? If so, would I just need to do another pull request for that draft?

mikeizbicki:
Yes, just issue another pull request.

PR #1598, opened 2015-05-20
added cs100-runtests README

alvtdev:
Added the README documentation for cs100-runtests.

Todo: 
- Obtain and add screenshots to the example walkthrough
- Possibly rewrite some parts to be less redundant and/or more informative
- Minor formatting changes.

### DIFF
--- a/textbook/tools/cs100-runtests/README.md
+++ b/textbook/tools/cs100-runtests/README.md
@@ -1,0 +1,158 @@
+#CS100 Runtests
+
+##Table of Contents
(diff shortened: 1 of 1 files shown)


alvtdev:
I have no questions so far, but if I do, I may ask them here or create a separate pull request with an updated readme.

I had written this under the assumption that it was for a separate repo, since I am working in one that contains runtests by itself, and completely forgot that the final product would actually be in gitlearn. I'll be sure to remove the "installation" section, as well as the "Features" section, since that doesn't pertain to demonstrating how to use runtests.

PR #1601, opened 2015-05-21
HW #4 Draft

baileyherms:
We will find ways to make sure it ties in with the assignments, right now we have descriptions and some examples.


mikeizbicki:
Any questions?

PR #1604, opened 2015-05-21
Hw4 rough 1st draft

knels012:
I haven't done much, but I wanted to see if the bare bones were okay. I listed the planned commands for the bash cheatsheet, making sure not to overlap the bash cheatsheet that already exists. I did a coupe of them to show how I plan to explain and show examples of the commands.


mikeizbicki:
You need to add onto the existing document, not create a new one.  Otherwise it looks reasonable.

PR #1596, opened 2015-05-20
HW4 First Draft

fvera002:

Issue #1487, opened 2015-05-05
HW 4 Idea


isadoramariamendes:
Hello Mike,
First of all, I would like to know if my partner and I need to post the idea or I just need to mention his name here.
Secondly, we were discussing about some ideas... and the one the we liked the most was basically to implement the Huffman Code using binary tree and dynamic heap. This is an interesting application of trees and discussing with some students I notice that most of them never heard about this method of lossless data compression... Researching in the internet I could not find the code implemented like this with heap and well documented. We believe this can be a nice project because most of new students in CS sometimes have the feeling that they are learning things that are not applicable (I heard things like "Why should I struct my project if I can put everything in the main file?" or "Why should I learn trees? I will never use it"... and this can be one example that will show them the opposite. Finally, my friend and I never implemented a dynamic heap and I can tell that I have some "good issues" implementing trees in general, so it's gonna be a challenge and at the final we will have learned a lot, as well. 

So at the final of the project we would have:
- BinaryTree.c and BinaryTree.h;
- DynamicHeap.c and DynamicHeap.h
- Huffman.c and Huffman.h
- Examples.c
- Documentation
- Explanations 

What do you think?

mikeizbicki:
You only need to post the issue once including both names.

About the idea, this would be a good project for a data structures class, but not for this class.  Huffman codes are a classic data structures problem and included in the standard libraries for most programming languages.  See the [boost library implementation](http://www.boost.org/doc/libs/1_35_0/libs/iostreams/doc/classes/zlib.html).  These existing implementations are already fully featured and highly optimized, so no one will want to use your library.

mikeizbicki:
There are a number of ideas related to this idea that might work.  But you'll need to emphasize: why does the world _need_ what you're making?

isadoramariamendes:
Can you give me one example that is related to this idea that might work? hahaha, now all ideas that I can think sounds "too data structures class" for me hahaha... =/

mikeizbicki:
You could create a short tutorial/lab that would teach people how to use
the existing zlib library and why it's useful.  This would help the same
group of people you were targeting with your initial proposal (i.e. people
who want to add compression to their programs), but it would do so in an
even better way by building upon existing work.

isadoramariamendes:
Hmmm, I took a look in this project here in Github and it is pretty well documented :( I don't think it would really help somebody...
[...]

PR #1606, opened 2015-05-21
HW 4 - 1st Draft

isadoramariamendes:
Replacing the 1st pull request because it had many commits...
https://github.com/mikeizbicki/ucr-cs100/pull/1596


fvera002:
Our write up is a focus is on LaTeX; however, we want to combine lots of unix tools together as suggested [here](https://github.com/mikeizbicki/ucr-cs100/issues/1487).

PR #1618, opened 2015-05-25
added video

zzhou007:
Links documentation for hw4

### DIFF
--- a/videos/links.md
+++ b/videos/links.md
@@ -1,0 +1,229 @@
+# Links
+###What is This?
+This is a page all about links for people who dont want to read the man pages or want examples.
+###What are Links and Why Should I Care?
+Without links Inode are associated with one directory entry at a time.
+With links, you can access the same inode with more than one entry.
+Why whould someone want to do this?
+Lets say your root partition is running out of space and you really want to instsall Libre office.
+You link a root folder to your home folder in another partition and install Libre Office in your home folder. 
+
+###Whats the Difference Between Hard and Soft Links?
+The main difference is soft links makes a new inode while hard links share the same inode.
+Click [Here](www.youtube.com/watch?v=j_BgOf2Ti-s) to watch a video about inodes.
+#Code 
+Dont want to read pages of test?
+Click [Here](https://www.youtube.com/watch?v=LPIMLR4simU) to watch a simplified video about the basics of links.
+###Creating Links
+So how whould I go about making these wonderful links?
+You can make hard links simply with the ln command.
+The format is 
+
+> ln file1 link1
+
+or
+
+> ln /path/to/file1 /path/to/link1
+
+Just make sure the file exists and the link does not exits.
+The command creates a hard link to file1.
+
+Soft links are exactly the same except with a '-s' flag.
+
+> ln -s file link
+
+or
+
+> ln -s /path/to/file1 /path/to/link1
+
+Soft links can link across file systems and to directories.
+###Removing Links
+Say you typed 'ln file1 lnik' instead of 'ln file1 link1'.
+To delete links, you can use
+
+> rm lnik
+
+or 
+
+> unlink lnik 
+
+Because files are directory entries associated with a Inode, you can use unlink to delete files.
+So whats the difference between the two?
+
+```
+$ touch $(seq 1 100) #makes file form 1 to 100
+$ time for i in $(seq 1 100); do rm $i; done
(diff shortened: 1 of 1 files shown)


zzhou007:
The videos still have not changed yet I just added documentation

mikeizbicki:
Any questions?

zzhou007:
So doing all the changes would an acceptable format for the write up be 

> the beginning portion on just creating hard links (with typos fixed)
> talk about -d flag with hard links and why it does not work 
> give an example of using a soft link inside drop box to back up files
> give an example of using soft links to link your home partition to your root partition of install programs so the root partition does not run out of space?

mikeizbicki:
You need to focus on coming up with good examples.  I can't say anything more specific about what is or isn't acceptable until I see the examples.  I'd expect to see at least 5 good examples on a writeup like this.

Examples you list:

>  give an example of using a soft link inside drop box to back up files

This is a good example, except that it depends on dropbox.  It will be hard to do this example without explaining how to install dropbox, which I don't want you to do.  You can use it, but you'll have to be careful.

> give an example of using soft links to link your home partition to your root partition of install programs so the root partition does not run out of space?

I don't understand exactly what you mean by this example.  There's probably something here, but you'll have to be a lot more explicit for me to approve it.

zzhou007:
For the first idea what if I replaced drop box with a cloud storage so it would just be syncing a folder to a cloud storage?

and changing the second idea what about an example of accessing and changing data from different OSs 
for example 
saving data such as picture documents and hw into a separate data partition and using soft links to access the data from both linux and windows?

Can I give these two examples to show the usefulness of links and make two videos on just the general idea and differences between hard and soft links and data blocks as a hybrid write up/video?

mikeizbicki:
> For the first idea what if I replaced drop box with a cloud storage so it would just be syncing a folder to a cloud storage?

It's better to be as specific as possible.  Doing dropbox is fine, just don't get distracted by explaining how to install dropbox.

> and changing the second idea what about an example of accessing and changing data from different OSs
> for example saving data such as picture documents and hw into a separate data partition and using soft links to access the data from both linux and windows?

That sounds perfect!

> Can I give these two examples to show the usefulness of links and make two videos on just the general idea and differences between hard and soft links and data blocks as a hybrid write up/video?

Quality is much more important than quantity.  I think you should focus a great job on either just a writeup, or just a video, and not both.  My recommendation is writeup because more people will read it and it will take less time.  But you can do videos if you want.

Either way, you will need to come up with more examples.  But those examples are great.

zzhou007:
So instead of doing videos what about a large write up of 
links 
link to another write up about inodes and data blocks as mentioned before 
differences between hard and soft as mentioned before 
section about why you cannot hardlink directories 
examples 1 and 2
example 3 of using hard links to organize your pictures folder to save space 
say you have thousands of photos and the photos are sorted into folder of that the picture depicts.
one folder is used to store pictures take during the summer 
the other folder is used to sure pictures of family 
if you have pictures of a family member taken during summer instead of making a copy of the file you can just use a hard link so you have the picture available in both folders and saving space 

example 4 
using hardlinks to make backups of very large files so if you accidentally run the rm command you still have the file. using hardlinks as a shortcut so you can access files on your desktop instead of moving through a bunch of directories 

example 5 
using soft links to run a program that constantly gets updated and has bugs and compatability issues so you need to keep multiple version 
ie 
program 2.04.2
program 2.04.1
you can use soft links called program pointing to the working version so you don't have to type as much

mikeizbicki:
Example 5 is great.  I recommend showing how a few real programs do this in
practice.  Git and gcc would be really good examples.  Git is relatively
simple because there's only a single program; gcc is relatively complicated
because there's multiple binaries and a few libraries.

I don't like example 4.

Example 3 seems okay, but I don't see why you are using hard links instead
of soft links?

Maybe you could replace example 4 with an example of how to "undelete"
something by recreating a link that you've removed?

zzhou007:
for example 3 i am using hard links to save more space given the picture are all on the same partition 

for example 4 i will try to write about how to undelete items given with debugfs 

thank you for the input,

mikeizbicki:
It's true that hard links useless space, but you'd have to have 1000s of hard links per image before this actually made a difference.  So I don't think this is a compelling reason for hard links.  A better reason would be so that you can move the files around without worrying about breaking all the soft links.

PR #1624, opened 2015-05-26
checking project progress

ghost:
checking the progress with you :
~one problem with the scripts is that we have issues with write permissions for files. If we do the path manually, it would save the files in the appropriate folders.


mikeizbicki:
> one problem with the scripts is that we have issues with write permissions for files. If we do the path manually, it would save the files in the appropriate folders.

I don't understand what this is trying to say.  But if you're having problems with the scripts, it's probably best to discuss them in person so that I can look at everything.

PR #1639, opened 2015-05-29
Squashed commit

cbarb005:
Squashed all into one commit. There's an issue with the ioccc-hamano folder in the Appendix. While it shows that the files are there, and git tells me everything is updated, once pushed, the files are missing. I tried manually re-adding two of them, and they appear just fine now, but I'm not sure how to handle the pdf files.


cbarb005:
Sorry we weren't able to talk on Friday, I had a meeting at noon on Friday to get to and there was quite a few people in line to talk with you already. Is Monday fine?

PR #1641, opened 2015-05-29
Record an enlightening sentence

hzhu007:
I just read it in a discussion. 
It is helpful to me and will also be helpful to future students i think.


mikeizbicki:
squash this into a single commit, and change `__PS__` to `**important:**` like everywhere else; then I'll merge

PR #1644, opened 2015-05-29
add general case and reformat

hzhu007:
Squashed ...


hzhu007:
To confirm, I am supposed to change **P.S.** to **IMPORTANT** and not to use a new line here right?

PR #1688, opened 2015-06-05
HW4 Submission

donei003:
I have added some useful gdb commands to the cheatsheet and did some minor changes to the existing cheatsheet. Additionally, I converted the old cheatsheet file to use the proper style guide in markdown where each sentence is on its own line.


mikeizbicki:
I need the file to be in the same location as the previous `gdb-cheatsheet.md` so that I can see the diff of changes you've made.

I'll also need the commits squashed into a single commit.

donei003:
Ok I will put them in the same location. But, how do you squash the commits into one?

PR #1693, opened 2015-06-06
HW4 Submission.

alvtdev:
I've reformatted my cs100-runtests readme such that the walkthrough is the first thing the reader sees, and includes screenshots. I've also included example test case and grade files for users to reference when creating their own. There is a small section with slightly more detailed explanations of each command after the example walkthrough.

In regards to not having the script default to bin/rshell, that would be an issue with the script itself, and I'm not proficient enough in bash scripting to change that. I've left the default in the documentation for the time being.


mikeizbicki:
I want this submited to the gitlearn repo.

PR #1697, opened 2015-06-06
hw4 submission - vim tutorial

kkita004:
final submission
I've added normal mode command keystrokes to the images and better integrated them with the text.
I've also moved the commands and their effects into tables.


mikeizbicki:
Also, your final submission needs to be a single commit.

PR #1699, opened 2015-06-06
second draft hw4

hzhu007:
Hi,

Here's the second draft of our hw4.


hzhu007:
@Renee7547 is my teammate.

PR #1713, opened 2015-06-08
HW4

jpatu001:


jpatu001:
I'm sorry for the bad writeup. Can I ask how much score I have with this right now? And do you have any advice on how I should focus the example? Thanks.

mikeizbicki:
As-is would be 10/50 on the technical section and 20/50 on the written section.  My advice is to focus the question: why should I care?  Basically, I can't give you an A in the class unless your writeup answers that question.

jpatu001:
Would it be better to just focus on `fmt` instead? Like do a lot of short examples explaining the use of each flag instead of actually going over the sort shuff and comm?

mikeizbicki:
Your question doesn't address the "why should I care?" problem, so I can't really answer it.  You just need real world examples that people would find out in the wild.

PR #1718, opened 2015-06-08
squashed commits

htang014:
Still interested in any extra credit opportunities


htang014:
We made the mistake of not working within the AWK branch, so squashing became a real problem real quick.  I hope this works.

PR #1719, opened 2015-06-08
HW4 Resubmit for vchan019 and htang014; Commits squashed

htang014:
We made the mistake of not working within the AWK branch so squashing became a problem.  I think it should be fine now.  Also were still interested in any extra credit opportunities


htang014:
I just realized some more extraneous file changes popped up.  Could you overlook this just this once?  We'll get it sorted out for the final draft

htang014:
Sorry Mike.  I don't see the lines that your are referring to in your comments.  Is there something Im missing?

htang014:
Thanks for the feedback Mike.  Just to clarify, in terms of completeness, have we covered our bases?  

And as far as EC goes...

PR #1736, opened 2015-06-09
Hw4 

TheBoop:
By David Ding(TheBoop) and Jerome Barbero(jbarb011)


mikeizbicki:
You need to include the actual images.  I can't give you feedback on the image quality until you do this.  I want them rendered as png files with a transparent background and 800 px wide.  The height can vary depending on the contents.

You also need to squash the commits into a single commit.

PR #1738, opened 2015-06-09
Textbook #2

cbarb005:
This version has added multiple READMEs as well as a few FAQ sections. I had some difficulty with the FAQs, since later homework issues tended to be more unique/specific than in the beginning.


mikeizbicki:
Overall this is excellent.  Thanks!  Make the changes I listed above and you'll get 110/100.

For your final submission, you'll need to squash your changes into a single commit, and you'll need to ensure that I don't have any merge conflicts.

PR #1742, opened 2015-06-09
Tor Relay Setup Guide

jpatu001:
Hey Mike, I had the messy GNU writeup that has no direction. I couldn't think of a good real world problem to get that topic fixed. So here I tried to salvage a really short guide with the Tor Relay. If this is worse than my previous pull request, I'll just take the 30 points that you'll give me for that previous  pull as is since I couldn't really think of anything else. Thanks


mikeizbicki:
There's a few technical errors in here, but it looks overall correct and was a great idea :)

I'm giving you an additional 15 points for this on top of the previous 30.

PR #1743, opened 2015-06-09
third submission

hzhu007:


hzhu007:
My teammate @renee7547 and I still have two final exams and don't have time to rework. This would be the final version.

mikeizbicki:
The technical information is good, but the writing could use a bit more polish.  85/100

Renee7547:
Hi Mike, could you please tell us the grade we got? Thank you.

Renee7547:
Oh, never mind, I got it.

PR #1744, opened 2015-06-09
hw4 should fit specs

TheBoop:
By David Ding(TheBoop) and Jerome Barbero(jbarb011)


mikeizbicki:
Overall, these images are VERY attractive looking.  Great job!

There's still a few minor edits though:

---

You need to delete the LICENSE file.

Remove the titles from the images.  When they get inserted into a writeup they'll look awkward.

When you call `open`, the file descriptor should be called `fd`, not `newfile`.

Latex is rendering your quotation marks wrong.  There's a number of different types of quotation marks.  You need the ones that go straight up and down.

Multiple forks image should call `wait` for each child.

Multiple forks with piping image: the stuff in the top box all happens within the parent, so should be moved into the green box.

Multiple forks with piping is missing some syscalls.  In particular, each child process must call `close` on the file descriptors it's not using anymore (that is, all the pipes that aren't involved with that process).  There will be a lot of syscalls, so 

All of the images with the green parent process should have a flow of time arrow coming into the first fork and out of the last wait.

The way that you handle flow of information in each of the io redirection/piping images is weird to me.  I want to see you put one of the file descriptor tables right below each child, and connect those tables together with information flow.  Don't forget to include the tty links in this as well.  The file `Piping.png` that shows just two processes is fine as-is, you don't need to do this to that file.

---

Make all these changes and you'll get 120/100

PR #1748, opened 2015-06-09
HW 4 submission

amah001:
redid GDB ex4 README.md along with strace video tutorial.


amah001:
My partner is ecar007, and we would not have time to work on this anymore so we would like it be graded as is.

mikeizbicki:
Sorry for the delay in grading.  I needed to find a computer with sound to watch the videos.

The videos themselves are rather long and don't tell much of a story.  They seem like they are more like "video manpages" than they are actual tutorials.  There are also a number of presentation mistakes in your videos.  Your written section on gdb is pretty good.

Overall, I'm giving you 75/100.

PR #1752, opened 2015-06-09
HW4 Resubmission

donei003:
I have moved the cheatsheet into the cheatsheet folder.


mikeizbicki:
Your commits still aren't correct.  Git thinks that you wrote the entire file, not that you're adding to someone else's file.  I need to be able to see the diff between your contributions and the already existing file to know what exactly you added.

You will also have to squash your commits into a single commit.  There's a good chance this will solve the other problem as well.

donei003:
The reason why it is doing that is because before, the writing had whole paragraphs on one "line". I fixed this and made it easier to read so that may add to some of that.

PR #1764, opened 2015-06-10
HW4 Submission

sthak004:
Hey Mike, 

I didn't have enough time to remove the previous commits that will be a part of this submission. i will fix it, but I just wanted to turn in the HW4 before my deadline.


mikeizbicki:
You need to squash your commits into a single commit before I can look at this.

PR #1765, opened 2015-06-10
HW4 Final Commit

skama004:
Just squashed all previous commits. My partners @divyanshch @hannesphillips


mikeizbicki:
I need this squashed into a single commit before I can look at it.

PR #1776, opened 2015-06-10
Hw 4 submission round 3.

jkinc002:
I think this should be the last one, even if there are errors in it. If it somehow comes out to less than the 70 points you said the previous version was worth, then I'll just submit that version. I can't spend anymore time on this.


mikeizbicki:
75/100

jkinc002:
Thanks for teaching.

PR #1778, opened 2015-06-11
hw4 submit with fixes

phung001:
This is the hw4 resubmission with the fixes that you suggested.


cdixi001:
from #1759

mikeizbicki:
Thanks for making those changes.  It is much better now.  But I still wouldn't be able to understand your explanations of io redirection if I didn't already know what you were talking about.  I'm upgrading your grades to 80/100.

Issue #1519, opened 2015-05-07
hw4-contributing proposal


mchen046:
@ykamo001 @ttruo018 and I, @mchen046, would like to create a utility that takes a git repo as a command line argument and outputs the structure of branches and respective commit history for each file in a pdf or PNG format. This will be done using LaTeX.

mikeizbicki:
Sounds good.  I'll want your program to be written as a bash shell script.  I suspect it'll be a fairly easy program to write.  The key will be to provide really excellent documentation for it.  That's what's going to make people want to use the project.  There's also a number of pictures of repositories in the various writeups inside this repo.  They're all in slightly different formats though, so I'll want you to recreate all these pictures using your tool.

mchen046:
Hey Mike, we currently have our project hosted on a github repository, hope that's fine: `https://github.com/ykamo001/GitMap`

So far, our project extracts all the information of all branches in a repository and stores the information into separate files. We are still working on implementing repository specification in the command line arguments. We also need to utilize the necessary information from the files to convert into a graph in LaTeX.

Currently we are implementing the tool in C++, though we still plan to implement the code in a bash shell script as per your request.

PR #1784, opened 2015-06-11
hw4 submission - GitMap

mchen046:
I apologize in advance, there are some rogue commits that appeared earlier in the quarter. What is the best course of action to remove the commit messages in this case?

Again, this is for I, @mchen046, @ttruo018, and @ykamo001.


mchen046:
Here's a link to the git repo: https://github.com/ykamo001/GitMap

mikeizbicki:
We should talk about this in person.  Can you all make my office hours tomorrow?

ykamo001:
I live in the By Area, so I won't be in Riverside until June 22 for summer session, sorry. Won't be able to come in for office hours.

mchen046:
I'm currently living 50 miles away from riverside, but if it's absolutely necessary I'll make the drive over there tomorrow. 

@ttruo018 says he'll be able to make it to your office hours tomorrow.

mikeizbicki:
You did not do what you proposed in #1519.  In particular, you said you would create a map of the repo using latex, and the output would be an image.  This would be very useful.  For example, I could automate the creation of all the repo images used in the textbook folder.

But if I understand your writeup, then all you're doing is printing the information from `git log` but visualized on the terminal in a slightly different way.  I don't see what the contribution is because I can get the exact same effect with a one line git command.  The `--pretty` parameter lets me specify in detail how I want each commit to be displayed, and I could use this parameter instead of your bash script to format the output the way you have.

Is the above correct?

mchen046:
Correct. We abandoned our original idea of using LaTeX. While we were writing the shell script, we extracted information using `git branch` and `git log`. From there, we decided to create dynamic box shapes using a mix of ascii characters and labeled the most recent commit of each branch. At the time, we were not aware of the `--pretty` parameter.

If you would like us to generate the .tex file that can then be converted into a pdf, we will write it as soon as possible.

mikeizbicki:
As-is, I'm struggling to find ways to give you any points.  If you can come up with a _specific_ plan on how you will fix the project and how long it will take, I'll _consider_ entering grade delays for you and upgrading your grade when you complete the project.  The final project would have to be _amazing_ and will suffer a significant penalty for being late.

PR #1792, opened 2015-06-11
HW4

tortas:
Sorry for the late squashing. I had no idea what you were talking about and it took a while to figure it out after I found the email update for this repo.


mikeizbicki:
I wish we would have talked about this more during the quarter.  Your contribution seems very minor.  It looks like you've added basically no new technical information.  Plus, your writing is too wordy and occasionally violates the styleguides.  I'm giving you 30/100.  If you think that I missed something, or that this is unfair, please let me know.

Your commit should be renaming the existing `cpp` files, not creating new versions.  This is important as it allows git to track who the author was.  You also need to delete the previous README files.  I can't merge your commit until this is complete.  Your grade will be entered when you finish this.

PR #1802, opened 2015-06-13
HW4 Final Submission

vhall415:


mikeizbicki:
Overall, this is very much improved both in terms of writing quality and examples.  There's still a few minor technical errors, so I'm giving you 40/50 on the technical portion.  There's also a few times when you're breaking the style guide, so you get 35/50 on the writing portion.  Plus the extra credit gives you 80/100 overall.

PR #1817, opened 2015-09-20
added MOSS

genchang1234:
Let me know what you think of the finished version.


mikeizbicki:
Because the images are so different than the actual code samples you have, I can't use this as is.  Therefore, I can't merge it, so we can't complete the agreement we had before.  

I'm sorry, about that, and I appreciate the work you've put into this.  But this has eaten up too much of my time already, and so I can't help you anymore with this.

